PR #1, opened 2023-03-19
ONUR-Pricing page exercise

Onuryemez54:


onur/styles.css
```diff
@@ -0,0 +1,241 @@
...
+    align-items: center;
+    gap: 1.2rem; 
+    list-style-type: none;
+}
+/* description section */
+.description {
+    background: (255, 255, 255);
+    max-width: 700px;
+    width: 70%;
+    height: 200px;
```
annasvst, line 64: I'd avoid setting height of this element explicitly as it may break on a different kind device. Instead, add enough spacing around the text (use margin or padding, which one is the most appropriate) and let the content take as much space as it needs to.

onur/styles.css
```diff
@@ -0,0 +1,241 @@
...
+    align-items: flex-start;
+    flex-wrap: wrap;
+}
+.prices {
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+}
+.head {
+    width: 18rem;
```
annasvst, line 105: I'd say if you set width explicitly for the card, you should only do so for the entire card (the most outer div of the card) and let all the other content be 100%. A better approach would be to align cards with CSS and not set the width explicitly at all (let it be as wide as content needs to) depending on the available space on the screen. It may sense to specify the min-width property though as it would make sure that the card does not get too small.

onur/styles.css
```diff
@@ -0,0 +1,241 @@
...
+    text-align: center;
+    border-radius: 4px 4px 0 0 ;
+}
+.head h1 {
+    font-weight: 100;
+    margin: 0.2rem auto;
+    font-size: 1.5rem;
+    padding: 0.3rem;
+}
+.body {
```
annasvst, line 119: Same here as about `.head`

onur/styles.css
```diff
@@ -0,0 +1,241 @@
...
+}
+.blue-btn {
+    background: rgb(30, 143, 218);
+    color: #fff;
+}
+.blue-btn:hover {
+    background: rgb(10, 103, 165);
+}
+/* footer section */
+.footers {
```
annasvst, line 159: It would be better to use a singular version of the word instead ('footer').

onur/styles.css
```diff
@@ -0,0 +1,241 @@
...
+.navbar ul li a:hover {
+    text-decoration:underline;
+    color: black;
+}
+
+.text-center {
+    text-align: center;
+}
+.heading-small {
+    font-size: 1.2rem;
```
annasvst, line 25: Nice that you decided to use 'rem'! It would be good to explicitly set the font-size in the `html` selector (e.g. 16px). Setting font size explicitly when using rem helps ensure a consistent, accessible, and flexible design that can adapt to different user preferences and devices (some browsers might have different default font sizes).

annasvst (commented):
Hi Onur, thank you for the great work! I've added some comments that could make the code a little bit better. :)

onur/index.html
```diff
@@ -1,13 +1,121 @@
...
+    <link href="https://fonts.googleapis.com/css2?family=Open+Sans:wght@400;600&display=swap" rel="stylesheet">
+    <!-- Font Awesome -->
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.3.0/css/all.min.css">
+    <link rel="stylesheet" href="./styles.css">
+    <title>Pricing page Exercise</title>
 </head>
 <body>
-  
+ <header class="conteiner">
+    <h3 class="heading-small">Company name</h1>
```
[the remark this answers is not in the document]
annasvst:
```suggestion
    <h3 class="heading-small">Company name</h3>
```

onur/styles.css
```diff
@@ -0,0 +1,241 @@
...
+    text-align: left;
+}
+
+.btn {
+    color:rgb(30, 143, 218);
+    border: 1px solid rgb(30, 143, 218);
+    border-radius: 5px;
+    padding: 0.6rem;    
+}
+.btn:hover {
```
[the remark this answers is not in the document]
Onuryemez54: thank you :)

PR #5, opened 2023-03-25
SALİH: pricing-page-exercise-v1

SalihBurakKus:


pricing-page-exercise/salih/index.html
```diff
@@ -1,13 +1,121 @@
...
+    <title>getbootstrap Pricing Page</title>
+    <link rel="stylesheet" href="styles.css">
 </head>
+
 <body>
-  
+    <nav>
+        <div class="nav-container">
+            <div class="company-name">
+                <span>Company Name</span>
```
annasvst, line 15: Usually, a link element (<a></a>) is used for a logo or company name in the navigation bar. This link should always take the user to the home page. I recommend to change <span></span> here to <a></a>.

pricing-page-exercise/salih/index.html
```diff
@@ -1,13 +1,121 @@
...
+                <div class="links">
+                    <ul class="menu" >
+                        <a href="#"><li class="menu-item">Features</li></a>
+                        <a href="#"><li class="menu-item">Enterprice</li></a>
+                        <a href="#"><li class="menu-item">Support</li></a>
+                        <a href="#"><li class="menu-item">Pricing</li></a>
+                    </ul>
+                </div>
+                <div class="sign-up-button">
+                    <a class="blue" href="#">Sign Up</a>
```
annasvst, line 27: I'd recommend to have this <a> element inside the unordered list above. You can keep it just as another item of the list. This would be more semantically correct and should not really affect styling as long as you add different classes to this element and other link elements in the list.

pricing-page-exercise/salih/styles.css
```diff
@@ -0,0 +1,195 @@
+/* Global */
+
+*, ::after, ::before {
+    box-sizing: border-box;
+    font-family:sans-serif;
```
annasvst, line 5: It's more common to add font-family to either `html` or `body` selector instead.

pricing-page-exercise/salih/styles.css
```diff
@@ -0,0 +1,195 @@
...
+}
+ul{
```
annasvst, line 7: It's ok to use `ul` selector in this case but it would be safer to use a class instead to select only the lists you need to change style for.

pricing-page-exercise/salih/styles.css
```diff
@@ -0,0 +1,195 @@
...
+    padding: 0;
+    text-align: center;
+}
+.main li{
+    padding: 0.2rem 0;
+}
+h1{
+    margin: 0;
+    text-align: center;
+    margin-top: 5px;
```
annasvst, line 18: I'd recommend combining margin into one property: e.g. `margin: 5px 0 0` or `margin: 5px 0 0 0` (they both mean the same thing).

pricing-page-exercise/salih/styles.css
```diff
@@ -0,0 +1,195 @@
...
+    margin: 15px 10px;
+    min-width: 60%;
+    padding: 12px 0;
+}
+small{
+    font-weight: 100;
+    font-size: 0.5em;
+}
+.blue-button{
+    background-color: #007bff;
```
annasvst, line 30: Try using only one type of colors (HEX is more commonly used).

pricing-page-exercise/salih/styles.css
```diff
@@ -0,0 +1,195 @@
...
+    flex-wrap: wrap;
+    border-bottom: 1px solid #dee2e6;
+    box-shadow: 0 0.25rem 0.75rem rgba(0, 0, 0, .05);
+}
+.menu-container{
+    display: flex;
+    align-items: center;
+    min-width: 35%;
+    justify-content: center;
+    /* border: 1px solid red; */
```
annasvst, line 65: It's better to always remove commented out code from your code however for the learning purposes it's ok to leave it in your personal projects.

pricing-page-exercise/salih/styles.css
```diff
@@ -0,0 +1,195 @@
...
+    list-style-type: none;
+    padding:0;
+    
+}
+.menu-item{
+    color: rgb(58, 58, 58);
+    padding: 1px;
+}
+.sign-up-button{
+    font-size: large;
```
annasvst, line 88: If you use  `em` as unit for font-size in other places, better to use it everywhere.

pricing-page-exercise/salih/styles.css
```diff
@@ -0,0 +1,195 @@
...
+    border: 1px solid rgb(92, 92, 255);
+    
+    padding: 6px;
+    border-radius: 2px;
+}
+
+.company-name{
+    margin-left: 1rem;
```
annasvst, line 96: It can be very confusing when different units are mixed in the same file/project. Pick one unit (`em` is used the most in your code)  and use it everywhere.

pricing-page-exercise/salih/styles.css
```diff
@@ -0,0 +1,195 @@
...
+/*______________________________________________*/
+/* The Cards are starting at this point */
+.main-container{
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+}
+.card-container{
+    width: 60%;
+    height: 50vh;
```
annasvst, line 132: Try not to set the height here as it may cause the layout to break on smaller screens when the content needs more space vertically

annasvst (commented):
Hi Salih, the code looks good overall, well done! I left minor comments mostly regarding using one single unit in the entire project.

pricing-page-exercise/salih/styles.css
SalihBurakKus: I deleted the height value. Then footer got closer to card-container. I added a margin-botton to card-container.

pricing-page-exercise/salih/styles.css
SalihBurakKus: I am using `rem `for spaces such as margin, padding, border, etc.
Prefer `em `for the fonts. But if it not right, I could move with one.

pricing-page-exercise/salih/styles.css
SalihBurakKus: I found two more and changed them also.

pricing-page-exercise/salih/styles.css
SalihBurakKus: I have found a few more and deleted them.

pricing-page-exercise/salih/styles.css
SalihBurakKus: I changed to all HEX code except the one which need to be transparent.

pricing-page-exercise/salih/styles.css
SalihBurakKus: Done

pricing-page-exercise/salih/styles.css
SalihBurakKus: I moved to `html`.

pricing-page-exercise/salih/index.html
SalihBurakKus: Done. I also changed `.links` css code to adapt this changement.

pricing-page-exercise/salih/index.html
SalihBurakKus: Okay, i made it `a `element with a link.

PR #6, opened 2023-03-25
DILARA - Pricing page exercise

dssanli:


pricing-page-exercise/dilara/styles.css
```diff
@@ -0,0 +1,234 @@
+.nav-bar {
```
annasvst, line 1: First thing, try to always include this code at the very top of your css file:

```

* {
  box-sizing: border-box; 
}
```

This would help avoid unexpected behaviour with element sizing. See Lesson 3 notes: https://classroom.google.com/c/NTg0NzkwMzQ4MjQ2/p/NTkyOTU1NjY3NTEz/details

pricing-page-exercise/dilara/styles.css
```diff
@@ -0,0 +1,234 @@
...
+    display: flex;
+    flex-wrap: wrap;
+    justify-content:space-between;
+    min-height:38px;
+    padding: 8px 24px;
+    align-items: center;
+    border-bottom: 1px solid #dee2e6;
```
annasvst, line 8: I see you mixing all kinds of colors. Try to use HEX colors everywhere and RGBA when you need to have a color with opacity. This is what's used the most in real life.

pricing-page-exercise/dilara/styles.css
```diff
@@ -0,0 +1,234 @@
...
+    font-size: medium;
+}
+
+.sign-up:hover {
+    background-color: #007bff;
+    color: white;
+}
+
+body {
+display: block;
```
annasvst, line 45: I think it should be `block` by default, so no need to set it explicitly like this.

annasvst (commented):
Good job Dilara!
Please see suggestions both in these comment and directly in the code. After implementing them, your code will be more optimized and maintainable.

In your code, you've used different units for font-size (medium, large, and px). It's recommended to use consistent units (e.g., px, em, or rem) to ensure a cohesive design across the site.

In several classes, you have the display: block; property. This is the default display property for most elements, so you can remove it unless you specifically need to override another display property applied earlier.

If you have common styles across multiple classes, you can create a separate class for those styles and apply them to the respective HTML elements to reduce redundancy.

For example, the .free, .pro, and .enterprise classes have some common styles. You can create a new class called .plan and apply it to the respective elements.

After implementing these suggestions, your CSS code will be even more optimized and maintainable.

pricing-page-exercise/dilara/index.html
```diff
@@ -8,6 +8,101 @@
...
+        <button class="second-button">Contact us</button>
+      </div>
+    </div>
+  </div>
+  <hr>
+  <div class="footer">
+    <div class="icon">
+      <img src="https://getbootstrap.com/docs/4.0/assets/brand/bootstrap-solid.svg" alt="" width="24" height="24">
+      <!--Add Icon Here-->
+      <p class="iconp">© 2017-2018</p>
```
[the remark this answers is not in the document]
dssanli: I couldn't change what you recommended to me. I left a paragraph because it shouldn't be the same line. Please correct me in this comment so I can see how I should do it.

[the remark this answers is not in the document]
annasvst: @dssanli all you need to do is to change html like this:

```
       <span class="iconp">© 2017-2018</span>
```

and add this line to CSS file to the .iconp class
```
display: block;
```

Paragraph is a block level element but span is an inline element by default. So when you switch to using <span> from using a <p>, you often need to set a display property to "block". :)

PR #14, opened 2023-04-03
Saffet-Responsive Navbar

Saffet01:


pricing-page-exercise/saffet/index.html
```diff
@@ -1,13 +1,138 @@
...
+  <title>Pricing example for Bootstrap (without Bootstrap)</title>
+  <link rel="icon" href="icon_path" type="image/icon type">
 </head>
+
 <body>
-  
+
+  <div class="nav-bar">
+
+    <p>Company name</p>
```
annasvst, line 17: I'd say it's more of a span element when you think about semantics. In reality, it would always be a link element on a page like this.

pricing-page-exercise/saffet/styles.css
```diff
@@ -0,0 +1,257 @@
...
+    border: 1px transparent grey;
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    padding-left: 25px;
+    padding-right: 35px;
+    box-shadow: rgba(99, 99, 99, 0.2) 0px 2px 8px 0px;
+}
+
+.nav-bar > p{
```
annasvst, line 21: I'd not recommend using a selector like this for this case. My suggestion is to make the "Company name" a link element to make it closer to what it would be in real life. Then it would be better to use id/class selector to add styles (my personal preference is to always use classes in CSS, not id).

pricing-page-exercise/saffet/styles.css
```diff
@@ -0,0 +1,257 @@
...
+    align-items: center;
+}
+
+.nav-links a{
+    text-decoration: none;
+}
+
+.nav-links>a{
+    padding-left: 25px;
+    color: black;
```
annasvst, line 36: Try to use HEX colors everywhere instead and RGBA when you need to have a color with opacity. This is what's used the most in real life.

pricing-page-exercise/saffet/styles.css
```diff
@@ -0,0 +1,257 @@
...
+}
+
+.button{
+    margin-left: 25px;
+    border: 1px solid #007bff;
+    padding: 10px 14px;
+    border-radius: 4px;
+}
+
+.button a{
```
annasvst, line 50: Does this code work? Maybe I'm missing something, but I cannot see any a elements inside of element with a "button" class in the html.

pricing-page-exercise/saffet/styles.css
```diff
@@ -0,0 +1,257 @@
...
+    color: #007bff;
+}
+
+.button:hover{
+    background-color: #007bff;
+}
+
+.button:hover a{
```
annasvst, line 58: Does this code work? Maybe I'm missing something, but I cannot see any a elements inside of element with a "button" class in the html.

pricing-page-exercise/saffet/styles.css
```diff
@@ -0,0 +1,257 @@
...
+
+.header{
+    font-size: 56px;
+    font-weight: 300;
+    margin-top: 70px;
+    margin-bottom: 10px;
+}
+
+.desc-text{
+    font-size: 1.2rem;
```
annasvst, line 77: I'd not recommend to mix different units, better to either stick to px, or rem, or em everywhere in the same project.

pricing-page-exercise/saffet/styles.css
```diff
@@ -0,0 +1,257 @@
+body {
+    font-family: sans-serif;
+    margin: 0;
+    padding: 0;
+}
+
+
+/* Navigation bar*/
+.nav-bar{
+    box-sizing: border-box;
```
annasvst, line 10: Always add box-sizing at the very beginning of the main CSS file to all elements at once, otherwise it will be very hard to debug unexpected behaviour. 
```
* { 
      box-sizing: border-box;
}
```

pricing-page-exercise/saffet/styles.css
```diff
@@ -0,0 +1,257 @@
...
+}
+
+.price span{
+    font-size: 1.4em;
+    font-weight: 500;
+    color: black;
+}
+
+.other-btn{
+    width: calc(100% - 40px);
```
annasvst, line 170: If you apply box-sizing, I think you'd not need to do this calculation, margin should be sufficient.

pricing-page-exercise/saffet/styles.css
```diff
@@ -0,0 +1,257 @@
...
+    line-height: 25px;
+}
+
+.footer-links a{
+    text-decoration: none;
+    color: grey;
+}
+
+.footer-links a:hover{
+    text-decoration: underline;
```
annasvst, line 236: You add the same hoover behaviour  to different element on the page. It would be nice to create a special class and use it for all the elements that need it.

annasvst (commented):
Well done Saffet, I left minor comments  but overall the code structure and semantics looks really good :)

pricing-page-exercise/saffet/index.html
Saffet01: "Company name" replaced with a <a> tag instead of <p>

pricing-page-exercise/saffet/styles.css
Saffet01: I used class to describe "Company name" as you adviced and I added <a> tag

pricing-page-exercise/saffet/styles.css
Saffet01: <div class="button">
        <a class="sign-up" href="#Sign-Up">Sign Up</a>
      </div>

In this part at the beginning, line 26. I used class named "button" and nest an <a> tag.
It changes "Sign Up" text color.

pricing-page-exercise/saffet/styles.css
Saffet01: It is same as just previous.

pricing-page-exercise/saffet/styles.css
Saffet01: I searched and found in web to save the day :/

pricing-page-exercise/saffet/styles.css
Saffet01: I generated from web so some lines has this situation but I will be more careful next time

pricing-page-exercise/saffet/index.html
```diff
@@ -1,13 +1,138 @@
...
+        <div class="card">
+          <h2 class="title-text">Free</h2>
+          <p class="price"><span>$0</span> / mo</p>
+          <ul>
+            <li>10 users included</li>
+            <li>2 GB of storage</li>
+            <li>Email support</li>
+            <li>Help center access</li>
+          </ul>
+          <button class="sign-btn">Sign up for free</button>
```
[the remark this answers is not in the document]
Saffet01: I changed it as an <a> tag but appearance changed also. I couldn't find the reason so I gave up on the change

pricing-page-exercise/saffet/styles.css
annasvst: Alright, I might missed it somehow :)

pricing-page-exercise/saffet/styles.css
annasvst: OK, leave it like this for now but it's always good to understand why the code works when you copy it. I know there's a lot to learn but trying to understand whatever solution you find will do you good in the long run :)

pricing-page-exercise/saffet/index.html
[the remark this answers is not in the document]
annasvst: Ok, keep it as is this time. However, for the links it's usually better to use <a> elements and then style it properly.

annasvst (approved):
Approved!

PR #16, opened 2023-04-08
Saffet responsive grid layout

Saffet01:


responsive-grid-layout/saffet/global.css
```diff
@@ -1,6 +1,7 @@
 * {
   box-sizing: border-box;
   font-family: "Segoe UI Adjusted", "Segoe UI", "Liberation Sans", sans-serif;
+  overflow: hidden;
```
annasvst, line 4: Why do you feel a need to specify overflow property in this case?

annasvst (approved):
Looks good Saffet, approved!

responsive-grid-layout/saffet/global.css
Saffet01: Because there was a scroll bar and I didn't want to see any scroll bar on right side

annasvst: Vertical scroll bar is expected in a layout like this as the page should be higher than the height of the screen so in this case `overflow: hidden;` is not relevant in this case.

PR #17, opened 2023-04-09
Dilara-grid-css

dssanli:
Updated with using grid template areas

### DIFF
--- a/responsive-grid-layout/dilara/global.css
+++ b/responsive-grid-layout/dilara/global.css
@@ -9,35 +9,6 @@ body {
   margin: 0;
 }
 
-.header {
-  background-color: #006c67;
-  display: grid;
-  grid-area: header;
-}
-.sidebar-left {
-  background-color: #003844;
-  grid-area: sidebar-left;
-}
-
-.sidebar-right {
-  background-color: #ffb100;
-  grid-area: sidebar-right;
-}
-.content {
-  background-color: #ffebc6;
-  grid-area: content;
-  
-}
-.footer {
-  background-color: #006c67;
-  grid-area: footer;
-  height: 200px;
-}
-
-.content > div {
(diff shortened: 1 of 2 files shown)


responsive-grid-layout/dilara/global.css
```diff
@@ -11,18 +11,27 @@ body {
 
 .header {
   background-color: #006c67;
+  display: grid;
```
annasvst, line 14: I don't think you need to make the div with 'header' class a gird container

responsive-grid-layout/dilara/global.css
```diff
@@ -11,18 +11,27 @@ body {
...
 }
 .content {
   background-color: #ffebc6;
+  grid-area: content;
+  
 }
 .footer {
   background-color: #006c67;
+  grid-area: footer;
+  height: 200px;
```
annasvst, line 34: I'm pretty sure that you don't need to set the height of the footer explicitly to 200px here. Since you don't change the default alignment (stretch), it will take as much space as you specify for the last row (300px in this case).

responsive-grid-layout/dilara/global.css
```diff
@@ -32,4 +41,10 @@ body {
 .container {
   width: 100vw;
   height: 100vh;
+  display: grid;
```
annasvst, line 44: I know that I placed classes initially in this order by my intention was not to use the classes above.
When doing your own projects, try to always place the code for container element above children elements. This would make your code more readable and help to avoid unexpected behaviour  related to specificity of elements (see more here: https://www.w3schools.com/css/css_specificity.asp).

annasvst (commented):
Well done, left minor comment / suggestion :)

annasvst (approved):
Approved!

PR #19, opened 2023-04-11
Grid-Layout-Tugba

TugbaMartin:


responsive-grid-layout/tugba/index.html
```diff
@@ -1,42 +1,19 @@
...
+<head>
+<meta charset="UTF-8">
+<meta name="viewport" content="width=device-width, initial-scale=1.0">
+<link rel="stylesheet" href="styles.css">
+<title>Grid Layout Exercise</title>
+</head>
+<body>
+<div class="container">
+<header>Header</header>
+<sidebar>sidebar-left</sidebar>
```
annasvst, line 12: I'd recommend to use a div instead, I haven't heard of a sidebar HTML element. 🙂

responsive-grid-layout/tugba/index.html
```diff
@@ -1,42 +1,19 @@
...
+<main>Content</main>
+<aside>Sidebar-right</aside>
```
annasvst, line 14: I think `aside` element may not be the most suitable element to use in this case. It's usually used for a smaller and less important content (see example here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside).

responsive-grid-layout/tugba/styles.css
```diff
@@ -0,0 +1,55 @@
+*{
+   margin: 0;
```
annasvst, line 2: I'd recommend to never set padding and margin to 0 for all elements (selector *)

responsive-grid-layout/tugba/styles.css
```diff
@@ -0,0 +1,55 @@
...
+    grid-template-columns: 1fr 4fr 1fr;
+    grid-template-rows: 80px 1fr 80px;
+    grid-template-areas:  
+    "header header header"
+    "sidebar-left content sidebar-right"
+    "sidebar-left footer footer"
+}
+
+header {
+    display: grid;
```
annasvst, line 25: For the sake of this exercise, I think you only need to set `display: grid;`for the div with class 'container'. All other elements don't really have contain that you'd need to align.

responsive-grid-layout/tugba/styles.css
```diff
@@ -0,0 +1,55 @@
...
+    background-color: aqua;  
+    grid-template-columns: 1 / span 5;
```
annasvst, line 27: `grid-template-columns: 1 / span 5;` is one of the ways to make the layout look as we need it to. However, using 'grid-area: header;'is a cleaner way to do that. You don't need to use both of them at the same time.

responsive-grid-layout/tugba/styles.css
```diff
@@ -0,0 +1,55 @@
...
+    grid-area: header;
+}
+
+sidebar{
+    display: grid;
```
annasvst, line 32: Same here about `display` and `grid-template-areas`.

responsive-grid-layout/tugba/styles.css
```diff
@@ -0,0 +1,55 @@
...
+   grid-template-areas: 1 / span 2;
+    grid-area: sidebar-left;
+    background-color: aquamarine;
+    }
+
+  aside {
+    display: grid;
```
annasvst, line 39: Same here about `display`.

responsive-grid-layout/tugba/styles.css
```diff
@@ -0,0 +1,55 @@
...
+  main {
+    display: grid;
+      grid-area: content;
+      background-color: paleturquoise;
+   
+      }
+footer{
+    display: grid;
+    background-color: skyblue;
+    grid-column: 2 / 4 ;
```
annasvst, line 53: You don't need to specify `display` and `grid-column` here.

annasvst (commented):
Hi Tuğba, good try! Left some comment with suggestions :)

responsive-grid-layout/tugba/index.html
TugbaMartin: I made ıt up and wanted to try. :) but I wıll change ıt. Thank you.

responsive-grid-layout/tugba/styles.css
TugbaMartin: I wıll definitely remember this note.

responsive-grid-layout/tugba/styles.css
annasvst: No worries, we all make mistakes, that's the only way to learn :)

annasvst (approved):
Approved!

PR #20, opened 2023-04-11
Pricing-Page-ONDER

ondermartin:


pricing-page-exercise/onder/index.html
```diff
@@ -4,10 +4,114 @@
...
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
-  <link rel="stylesheet" href="./styles.css">
+<link rel="stylesheet" href="./styles.css">
   <title>Pricing page Exercise</title>
 </head>
 <body>
-  
+ <header>
+    <nav class="navbar">
+        <div><h3 class="heading-small">Company name</h3></div>
```
annasvst, line 13: If this would be a real website, usual behaviour is when a user clicks on the logo/company name in the navbar, he/she is redirected to the main page. So using a link element (<a></a>) instead of <h3></h3> would be more appropriate.

pricing-page-exercise/onder/index.html
```diff
@@ -4,10 +4,114 @@
...
+                    <p class="small">© 2017-2018</p>
```
annasvst, line 80: Using a `span` element would be more semantically correct here as the text "© 2017-2018"  is not a paragraph or even a sentence.

pricing-page-exercise/onder/styles.css
```diff
@@ -0,0 +1,215 @@
...
+}
+
+.text-center {
+    text-align: center;
+  font-weight: bold;
+}
+
+
+.btn {
+    color:rgb(30, 143, 218);
```
annasvst, line 57: Try to only use one kind of colors in your project (usually either HEX or RGB is used, however HEX is more common). It's ok to mix HEX and RGBA which you need to provide the "A" value for opacity.

pricing-page-exercise/onder/styles.css
```diff
@@ -0,0 +1,215 @@
...
+    color:white;
+}
+.btn:active {
+    box-shadow: 0 0 2px 2px #3e8fc5;
+}
+
+
+/* description section */
+.description {
+    background: (255, 255, 255);
```
annasvst, line 74: I think you meant `rgb(255, 255, 255);` here

pricing-page-exercise/onder/styles.css
```diff
@@ -0,0 +1,215 @@
...
+    max-width: 700px;
+    width: 70%;
+    padding: 0.5rem;
+    margin-bottom: 4rem;
+    margin-top: 4rem;
+    margin-left: auto;
+    margin-right: auto;
```
annasvst, line 81: When you only need to set margin for some of the sides, using a shorthand property is common: `margin: 4rem auto;`

annasvst (commented):
Very well done Önder! I left some suggestion to improve the code however I think the structure and semantics are really good :)

annasvst (approved):
Approved!

PR #23, opened 2023-04-13
Pricing-Page-TUGBA

TugbaMartin:


pricing-page-exercise/tugba/index.html
```diff
@@ -4,10 +4,114 @@
...
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
-  <link rel="stylesheet" href="./styles.css">
+<link rel="stylesheet" href="./styles.css">
   <title>Pricing page Exercise</title>
 </head>
 <body>
-  
+ <header>
+    <nav class="navbar">
+        <div><h3 class="heading-small">Company name</h3></div>
```
annasvst, line 13: Usually, there's a logo or a company name in the navigation bar on every website. When you click on it, it always takes you to the home page. 
So in this case it would make more sense to use `<a>` element instead of `<h3>`.

pricing-page-exercise/tugba/index.html
```diff
@@ -4,10 +4,114 @@
...
+                    </ul>
+                    <button class="btn price-btn blue-btn">Get started</button>
+                </div>
+            </div>
+            <div class="prices">
+                <div class="card_head">
+                    <h1 class="text-center">Enterprise</h1>
+                </div>
+                <div class="card_body">
+                    <h1><span class="dollar">&#36;29</span><span>/ mo</span></h1>
```
annasvst, line 65: Same here about h1

pricing-page-exercise/tugba/styles.css
```diff
@@ -0,0 +1,215 @@
...
+    color:white;
+}
+.btn:active {
+    box-shadow: 0 0 2px 2px #3e8fc5;
+}
+
+
+/* description section */
+.description {
+    background: (255, 255, 255);
```
annasvst, line 74: You probably meant `background: rgb(255, 255, 255);`

pricing-page-exercise/tugba/styles.css
```diff
@@ -0,0 +1,215 @@
...
+.prices {
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+    min-width: 18rem;
+
+}
+.card_head {
+    width: 100%;
+    height: 3rem;
```
annasvst, line 122: You can of course specify height, width, min-width (as above) using rem but it may not be as easy to understand for the one reading your code. I'd stick to px, % or vh/vw for width and height instead. 

Also, you probably don't want to set height like this. It's usually better to add enough margin/padding and let the content of the div decided how much height it should take.

annasvst (commented):
Hi Tuğba, well done! 👏 I've added some comments with possible improvements to make your code event better :)

annasvst:
Closing this PR in favor of https://github.com/frontend-from-0/240123-exercises/pull/45#pullrequestreview-1620687927

PR #25, opened 2023-05-15
Onur-Lesson13-js-file

Onuryemez54:


JS/Lesson13/onur/script.js
```diff
@@ -94,36 +98,57 @@ console.log('------------------------');
 // The reason the output is one hour behind what you might expect (i.e. January 1, 2000 at 00:00:00) is because of the timezone offset between your local timezone and UTC. The output indicates that your local timezone is one hour ahead of UTC. This is likely because your local timezone observes daylight saving time, which can cause the timezone offset to change depending on the time of year.
 // To get the date and time in your local timezone, you can use methods such as getDate(), getMonth(), and getFullYear() to extract the individual components of the date and time, and then construct a new Date object using those components. 
 
+const TODAY_DATE = new Date("05-15-2023 03:02");
```
annasvst, line 101: You can also get todays date by using this code `const TODAY_DATE = new Date();`. In this case, a date will be generated when code runs while `new Date("05-15-2023 03:02");` always generates the same date that is specified in the parentheses.

annasvst (approved):
Well done Onur! Left minor comment about Date() function otherwise looks great! No need to fix it this time, just information for the future.

PR #27, opened 2023-05-19
Saffet-Lesson 13 Exercises

Saffet01:


JS/Lesson13/saffet/script.js
```diff
@@ -3,71 +3,64 @@
...
 console.log('------------------------');
+
 // ---
 // 3. Declare a variable number1 and initialize it to 10.
 // Declare a variable number2 and initialize it to 3.
 // Declare a variable result and initialize it to the quotient of number1 divided by number2.
 // Declare a variable number3 and initialize it to the difference between number1 and number2.
-const number1 = 10;
-const number2 = 3;
+var number1 = 10;
```
annasvst, line 30: It's better to use const or let in modern JS. Var keyword was the only keyword for defining variables in the past and it's still present in the language but you should be able to only use const and let.

annasvst (commented):
Well done Saffet! The only thing to be fixed is to replace `var` keywords with either `let`(if value is changing) or `const`

annasvst (approved):
Well done!

PR #36, opened 2023-06-12
Onur-Lesson 15 files

Onuryemez54:


JS/Lesson15/onur/script.js
```diff
@@ -121,10 +57,44 @@ checkIfStringIsUpperCase('write a program that takes in');
 
 // 5. Write a program that takes in a number and checks if it's positive, negative, or zero using if else statement.
 
+function isNumberPositiveOrNegative(num) {
```
annasvst, line 60: A note for the future, since JS is not strongly types language, it maybe that the value that is passed to the function is not even a number. In real world, we usually check that the value passed is a number first, if it's not, the function should not even do the checks.

JS/Lesson15/onur/script.js
```diff
@@ -121,10 +57,44 @@ checkIfStringIsUpperCase('write a program that takes in');
...
+	}
+}
+isNumberPositiveOrNegative(65);
+isNumberPositiveOrNegative(-27);
+isNumberPositiveOrNegative(0);
+
 // 6. Write a program that takes in a year and checks if it's a leap year (is divisible by 4 and not divisible by 100 OR divisible by 400) or not using if else statement.
 
+function isLeapYear(yearNumber) {
+	if (yearNumber % 4 === 0 && (yearNumber % 100 !== 0 || yearNumber % 400 === 0)) {
```
annasvst, line 76: Your solution works (as numbers divisible by 400 are also divisible by 4) but is harder to comprehend. I'd do like this instead: 
```suggestion
	if (yearNumber % 4 === 0 && yearNumber % 100 !== 0 ) || yearNumber % 400 === 0)) {
```
so it is either yearNumber divisible by 4 and not divisible by 100 OR yearNumber divisible by 400.

JS/Lesson15/onur/script.js
```diff
@@ -121,10 +57,44 @@ checkIfStringIsUpperCase('write a program that takes in');
...
+}
+isLeapYear(2000);
+isLeapYear(2023);
+
 // 7. Write a program that takes in a temperature and checks if it's above or below freezing using if else statement.
 
+function isTemperatureCheckFreezing(temperature) {
+	if (temperature <= 0) {
+		console.log('temperature is below freezing');
+	} else {
```
annasvst, line 90:
```suggestion
	} else if (temperature > 0) {
```

annasvst (approved):

PR #40, opened 2023-06-19
Saffet-lesson15

Saffet01:


JS/Lesson15/saffet/script.js
```diff
@@ -67,113 +67,348 @@ switch (day) {
...
+console.log("-------------------------------------------------");
+// 2. Write a program that takes age as input. If the age is less than 18, print "Sorry, you are not old enough to vote". 
+// 	  If the age is between 18 and 65 (inclusive), print "You are eligible to vote". If the age is greater than 65, print 
+// 	  "You are a senior citizen".
+console.log("Ex2:");
+function checkAge(age){
+	if(age < 18){
+		console.log("Sorry, you are not old enough to vote");
+	}
+	else if(age>18 && age<65){
```
annasvst, line 91: In this case, checkAge does not take into account numbers 18 and 65 (they both will return "You are a senior citizen") which is not expected. Include both numbers in the condition like this: `(age >= 18 && age <= 65)`

JS/Lesson15/saffet/script.js
```diff
@@ -67,113 +67,348 @@ switch (day) {
...
+checkSign(-4);
+checkSign(0);
+
+console.log("-------------------------------------------------");
+// 6. Write a program that takes in a year and checks if it's a leap year 
+//(is divisible by 4 and not divisible by 100 OR divisible by 400) or not using if else statement.
+console.log("Ex6:");
+
+function checkLeapYear(year){
+	if(year % 4 == 0 && year % 400 == 0){
```
annasvst, line 164: This condition is not correct. Please make sure that the number is **divisible by 4 and not divisible by 100**

JS/Lesson15/saffet/script.js
```diff
@@ -67,113 +67,348 @@ switch (day) {
...
+
+checkLeapYear(1800); // it is not leap year.
+checkLeapYear(1600); // it is a leap year.
 
+console.log("-------------------------------------------------");
 // 7. Write a program that takes in a temperature and checks if it's above or below freezing using if else statement.
+console.log("Ex7:");
 
+function checkTemp(temperature){
+	if(temperature <= 0){
```
annasvst, line 180: 0 is actually the freezing point so should be `temperature < 0`

JS/Lesson15/saffet/script.js
```diff
@@ -67,113 +67,348 @@ switch (day) {
...
 // 25.0 – 29.9	Overweight
 // 30.0 and Above	Obesity
+console.log("Ex8:");
+function chechBody(weight, height){
+	let bmi = weight / (height*height);
+
+	if (bmi < 18.5)
+	console.log("underweight");
 
+	else if (bmi >= 18.5 && bmi < 24.9)
```
annasvst, line 206: 24.9 should be included in the "Healthy Weight"

JS/Lesson15/saffet/script.js
```diff
@@ -67,113 +67,348 @@ switch (day) {
...
+	console.log("Healthy");
+
+	else if (bmi >= 24.9 && bmi < 30)
+   console.log("overweight");
```
annasvst, line 210: "Overweight" should start from 25

JS/Lesson15/saffet/script.js
```diff
@@ -67,113 +67,348 @@ switch (day) {
...
+
+chechBody(25, 1.69);
+chechBody(75, 1.69);
+chechBody(100, 1.69);
+
+
+console.log("-------------------------------------------------");
 // 9. Write a program that takes in a person's age and checks if they are a child (age 0-12), teenager (age 13-19), adult (age 20-64), or senior citizen (age 65 and above) using if else statement.
+console.log("Ex9:");
+function checkCategory(age1){
```
annasvst, line 224: You should not have to write age1 as a name of a function parameter. The parameter is local to the scope of the function so as long as there are no other parameter with name "age" that this function takes (and it does not) and no other variables with name "age" declared inside the function itself, it should be safe to name the parameter "age" instead of  "age1"

JS/Lesson15/saffet/script.js
```diff
@@ -67,113 +67,348 @@ switch (day) {
...
+	if(age1<12 && age1>0){
```
annasvst, line 225: This condition works but it seems to be inverted from the readability perspective. We read from left to write so it makes code cleaner when you put the smaller condition to the left, the bigger to the right: `age > 0 && age =< 12)`.
Also, make sure 12 is included in the condition!

JS/Lesson15/saffet/script.js
```diff
@@ -67,113 +67,348 @@ switch (day) {
...
+		console.log("Child");
+	}
+	else if(age1>13 && age1<19){
+		console.log("Teenager");
+	}
+	else if(age1>20 && age1<64){
```
annasvst, line 231:
```suggestion
	else if(age1>=20 && age1=<64){
```

JS/Lesson15/saffet/script.js
```diff
@@ -67,113 +67,348 @@ switch (day) {
...
+//If the age is between 18 and 65 (inclusive) and the user is not a student, print "You are eligible to vote". 
+//If the age is greater than 65, print "You are a senior citizen". If the user is a student but their age is greater than or equal to 18, print 
+//"You are old enough to vote".
+console.log("Ex10:");
+
+function fonk(age, studentFlag){
+	if(age<18 && studentFlag ==="yes"){
+		console.log("You are a student and not old enough to vote");
+	}
+	else if(age>18 && age<65 && studentFlag==="no"){
```
annasvst, line 255:
```suggestion
	else if(age>=18 && age=<65 && studentFlag==="no"){
```

JS/Lesson15/saffet/script.js
```diff
@@ -67,113 +67,348 @@ switch (day) {
...
+	else if(age>65){
+		console.log("You are a senior citizen");
+	}
+	else if(studentFlag ==="yes" && age >= 18){
+		console.log("You are old enough to vote")
+	}
+}
 
-// 11. Write a program that asks the user to enter their salary and their years of experience. If the salary is less than $30,000 and the years of experience are less than 5, print "Sorry, we cannot offer you the job at this time". If the salary is less than $50,000 and the years of experience are between 5 and 10 (inclusive), print "We can offer you the job at a lower salary". If the salary is greater than or equal to $50,000 or the years of experience are greater than or equal to 10, print "We can offer you the job at the requested salary".
+fonk(12, "yes");
```
annasvst, line 266: When you have a parameter that can only take two values (like yes/no), it's preferred to use a boolean value for it: studentFlag would be either true or false. The boolean cannot take and other value apart from these two so your code will also be safer.

JS/Lesson15/saffet/script.js
```diff
@@ -67,113 +67,348 @@ switch (day) {
...
-		case 11:
-			console.log('Autumn');
-			break;
-		default:
-			console.log('Invalid Input');
+function checkScore(score){
+	if(score>0 && score<60){
+		console.log("You failed the test");
+	}
+	else if(score>60 && score<70){
```
annasvst, line 305:
```suggestion
	else if(score>=60 && score=<70){
```

JS/Lesson15/saffet/script.js
```diff
@@ -67,113 +67,348 @@ switch (day) {
...
+		console.log("You passed the test, but your grade is not great");
+	}
+	else if(score>71 && score<80){
```
annasvst, line 308:
```suggestion
	else if(score>=71 && score=<80){
```

JS/Lesson15/saffet/script.js
```diff
@@ -67,113 +67,348 @@ switch (day) {
...
+		console.log("You got a B");
+	}
+	else if(score>81 && score<90){
```
annasvst, line 311:
```suggestion
	else if(score>=81 && score=<90){
```

JS/Lesson15/saffet/script.js
```diff
@@ -67,113 +67,348 @@ switch (day) {
...
+findGrade("A");
+findGrade("B");
+findGrade("F");
+findGrade("f");
+console.log("-------------------------------------------------");
+// 15. Write a program that asks the user to enter their age, and then calculates and displays their age in dog years (where one human year is equivalent to seven dog years). 
+//If the user enters a negative number or zero, display an error message.
+console.log("Ex15:");
+
+function dogYears(age2){
```
annasvst, line 401: Try to use verbs in the names of the functions as they always do some kind of action
```
function convertAgeToDogYears(age){
```

annasvst (changes requested):
Hi Saffet, well done overall. The most errors are connected to the equality in conditions, do you need any help with it?

JS/Lesson15/saffet/script.js
Saffet01: I understand, just I thought that there are different functions and I need to seperate these functions' parameters easily. Thanks

annasvst (approved):
Approved, please merge to master

PR #44, opened 2023-07-09
Saffet-Lesson19

Saffet01:


JS/Lesson19/saffet/exercises.js
```diff
@@ -1,8 +1,201 @@
...
+// 1. Create a Person class with name, age, and gender properties, and a greet() method that logs a greeting message to the console. 
+// Then create a Student class that extends the Person class and adds a major property. Override the greet() method in the Student 
+// class to log a different greeting message to the console. Create 2 variables, one of type Person and one of type Student and 
+// log them to console.
+console.log("Example 1; ");
 
-// 2. Create a Shape class with name and color properties, and a describe() method that logs a description of the shape to the console (e.g. "This is a blue square"). Then create a Square class that extends the Shape class and adds sideLength and area properties. Override the describe() method in the Square class to log a description of the square to the console (e.g. "This is a blue square with a side length of 5 and an area of 25").
+class Person {
+    constructor(name, age, gender){
+        this.name = name;
```
annasvst, line 9: It's always better to use "private values inside a constructor, e.g. `this._name = name` and then in the `greet` method you'd use `this._name`

JS/Lesson19/saffet/exercises.js
```diff
@@ -1,8 +1,201 @@
...
+    }
+
+    describe(){
+        console.log(`This is a ${this.color} ${this.name}`);
+    }
+}
+
+class Square extends Shape{
+    constructor(name = "Square", color, sideLenght){
+        super(color);
```
annasvst, line 56: You need to pass both `name` and `color` to the `super` method, without it you get "undefined" instead of the color in the message "This is a undefined Square with a side lenght of 4 and area of 16".

JS/Lesson19/saffet/exercises.js
```diff
@@ -1,8 +1,201 @@
...
+
+    describe(){
+        console.log(`This is a ${this.color} ${this.name} with a side lenght of ${this.sideLenght} and area of ${this.area}`);
+    }
+}
+
+const myShape = new Shape("rectangular", "purple");
+myShape.describe();
+
+const mySquare = new Square("Square", "blue", 4);
```
annasvst, line 70: Since `square` will always be a `square`, it does not need to take a `name` parameter, makes sense to hardcode it in the Square class.

JS/Lesson19/saffet/exercises.js
```diff
@@ -1,8 +1,201 @@
...
+// with properties for type (either "deposit" or "withdrawal"), amount, and date ({type: "withdrawal", amount: amount, 
+// date: new Date().toLocaleDateString()}). Implement deposit() and withdraw() methods that add new transactions to the 
+// transactionHistory array. Implement a get transactionHistory() getter method that returns a copy of the transaction history array, 
+// and a get currentBalance() getter method that calculates and returns the current balance of the account based on the 
+// transaction history.
+console.log("Exercise 4: ");
+
+class BankAccount{
+    constructor(balance, interestRate){
+        this. balance = balance;
```
annasvst, line 120: I don't this the code works with the space between `this.` and `balance`

JS/Lesson19/saffet/exercises.js
```diff
@@ -1,8 +1,201 @@
...
+        this.interestRate = interestRate;
```
annasvst, line 121: Make all these values private (use underscore), not only `transactionHistory`

JS/Lesson19/saffet/exercises.js
```diff
@@ -1,8 +1,201 @@
...
+                console.log("Insufficient balance.");
+            }
+        }
+        else{
+            console.log("Provide a positive value.");
+        }
+    }
+
+    get transactionHistory() {
+        return [...this._transactionHistory];
```
annasvst, line 163: Nice that you return a shallow copy of the data instead of the original, it's always good idea to protect the original values :)

JS/Lesson19/saffet/exercises.js
Saffet01: I don't know if I understand correctly or not. But I hope it is.

annasvst (approved):
Approved!

PR #46, opened 2023-07-16
Tugba martin JS14 updated

TugbaMartin:


JS/Lesson14/tugba/script.js
```diff
@@ -69,11 +112,18 @@ console.log('------------------------');
...
 // Question: What is the difference between `substring()` and `slice()`?
 const stringExercise14 = 'Databases are an essential component of many applications.';
+const extractedWord = stringExercise14.substring(0, 9);
+
+console.log("Extracted word:", extractedWord);
 
-console.log('------------------------');
 // 15. Remove extra spaces from the string stringExercise15 and print the result to the console.
 const stringExercise15 = "    Hello world    ";
+const trimmedString = stringExercise15.trim().replace(/\s+/g, ' ');
```
annasvst, line 121: `.trim()` should remove the spaces, you don't need to use both `trim()` and `.replace(/\s+/g, ' ')`

JS/Lesson14/tugba/script.js
```diff
@@ -69,11 +112,18 @@ console.log('------------------------');
...
+
+console.log("Trimmed string:", trimmedString);
 
-console.log('------------------------');
 // 16. Extract the price and currency from a string below using "slice" method and print it to console.
-const planPriceString = "Premium plan - 9.99 USD/month."; // Should return "9.99 USD"
+const planPriceString = "Premium plan - 9.99 USD/month.";
+const extractedPriceCurrency = planPriceString.slice(planPriceString.indexOf(" ") + 1, planPriceString.indexOf("/"));
```
annasvst, line 127: I don't think this works. `planPriceString.indexOf(" ") + 1` searches for the first occurance of a space and it's not what you need. Instead, you can look for "-" or even "- " (dash and a space) as it's the separator that's used before the price. :)

JS/Lesson14/tugba/script.js
annasvst, line 121:
```suggestion
const trimmedString = stringExercise15.trim();
```

annasvst (approved):
Approved!

PR #50, opened 2023-07-16
Tugba-Martin -JS-Lesson17

TugbaMartin:


JS/Lesson16/tugba/script.js
```diff
@@ -60,83 +60,137 @@ console.log(newArray);
...
-let randomNumber = getRandomNumber();
-let randomNumbers = [randomNumber];
-
-while(randomNumber <= 8){
-randomNumber = getRandomNumber();
-randomNumbers.push(randomNumber);
-iteration++;
+
+let totalIterationsNumber = 0;
+let randomNumbers = 0 [randomNumber];
```
annasvst, line 74: This is not a correct JS syntax. What were you trying to do here?

JS/Lesson16/tugba/script.js
```diff
@@ -60,83 +60,137 @@ console.log(newArray);
...
+  sum += i;
+}
+
+console.log("Sum of even numbers:", sum);
 
 
 console.log('- - - - - - - - - - - - - - - - - ');
 // 14. Write a loop that finds the second largest number in an array of numbers.
 const numbers14 = [10, 5, 20, 8, 15];
+let largest = -Infinity;
```
annasvst, line 152: The solution for this exercise does not seem to work.

Try to start like this:
```
    let firstLargest = Math.max(numbers[0], numbers[1]);
    let secondLargest = Math.min(numbers[0], numbers[1]);
```
You'll get two first numbers of the array an assign them two the variables depending on which one of them is bigger.
Then start a loop and compare the rest of the array with the values that you assigned already. You'll need to start from the 3rd number (index 2), not from 0.

JS/Lesson17/tugba/script.js
```diff
@@ -134,19 +134,54 @@ console.log('----------------------------');
 // Input: ({name: "John", age: 30}, "city", "New York")
 // Output: {name: "John", age: 30, city: "New York"}
 
+//I'm really not sure about 10-11-12-13 questions.
+const object = { name: "John", age: 30 };
```
annasvst, line 138:
```suggestion
const object = { name: "John", age: 30 };
object["city"] = "New York";
```
Here I used the bracket notation to add a new key to the object and assign a value to it

annasvst: You don't need the rest of the code except the consol.log statement for this exercise

JS/Lesson17/tugba/script.js
```diff
@@ -134,19 +134,54 @@ console.log('----------------------------');
...
 
+function removeObjectKey(obj, key) {
+    delete obj[key];
+    return obj;
+  }
+  const object = { name: "John", age: 30, city: "New York" };
+  const key = "city";
+  
+  const result = removeObjectKey(object, key);
+  console.log(result);
```
annasvst, line 160: This is a correct solution!

JS/Lesson17/tugba/script.js
```diff
@@ -134,19 +134,54 @@ console.log('----------------------------');
...
 console.log('----------------------------');
+
 // 12. Write a function that takes two arrays of equal length and returns an object with the first array as keys and the second array as values.
 // Input: (["name", "age", "city"], ["John", 30, "New York"])
 // Output: {name: "John", age: 30, city: "New York"}
 
+const keys = ["name", "age", "city"];
+const values = ["John", 30, "New York"];
+
+const object = arraysToObject(keys, values);
```
annasvst, line 171:
```
function arraysToObject(keys, values) {
    const result = {}; // Create a new empty object first
    for (let i = 0; i < keys.length; i++) { // Use loop to iterate through each of the values
        result[keys[i]] = values[i]; // Add the key value pair to the object
    }

    return result; 
}
```

annasvst (changes requested):
Hi Tuğba, both lesson 16 and 17 reviewed in this PR. You Tugba-Martin-JS-Lesson17 branch to add changes for both assignments. :)

PR #51, opened 2023-07-17
Saffet-Lesson18

Saffet01:
It was really difficult to understand every steps. I tried to complete all examples but for almost half of it I needed to search in internet. After all I am not still feeling prepared for another example. I hope we face this matter really rarely :(


annasvst:
I'm glad you did not gave up. 💪
Recursion is hard but the only way to make something hard easier is to practice more :) 

Please remove lesson17 file from the PR and I'll approve

annasvst (approved):
Approved

PR #54, opened 2023-07-30
lesson19-salih

SalihBurakKus:


JS/Lesson19/salih/exercises.js
```diff
@@ -1,8 +1,157 @@
...
+        this._name = name;
+        this._color = color;
+    }
+    describe(){
+        console.log("This is a " + this._color + " " + this._name);
+    }
+}
+
+class Square extends Shape{
+    constructor(name, color, sideLength, area){
```
annasvst, line 43: Square will always be a square, so you don't need to have a `name` parameter, you can instead hardcode it to "Square"

JS/Lesson19/salih/exercises.js
```diff
@@ -1,8 +1,157 @@
...
+        this._make = make;
+        this._model = model;
+        this._year = year;
+    }
+    start(){
+        console.log("The " + this._model + " model vehicle has started.");
+    }
+}
+
+class car extends Vehicle{
```
annasvst, line 74:
```suggestion
class Car extends Vehicle{
```

annasvst (changes requested):
Good job Salih, left minor comments :)

annasvst (approved):
Approved!

PR #55, opened 2023-08-04
salih-lesson16-js

SalihBurakKus:


JS/Lesson16/salih/script.js
```diff
@@ -116,27 +116,91 @@ for(const key in obj) {
...
+		big = firstNum;
+	} else {
+		small = firstNum;
+		big = secondNum;
+	};
+	console.log('11-The Small Number is ', small, '. The Big Number is ', big); //I used them to follow the code.
+	for (i=small; i < big+1; i++){
+		console.log('22-The Small Number is ', small, '. The Big Number is ', big, 'The i Number is ', i); //I used them to follow the code.
+		if (i % 2 == 0){
+			sum += small;
```
annasvst, line 144: I think the mistake is here. You need to use `i` as you're trying to loop trough small to the big number (including).

JS/Lesson16/salih/script.js
```diff
@@ -116,27 +116,91 @@ for(const key in obj) {
...
+}
 
-
-
+console.log('Exercise 13.1: ', sumEvenNumbers(1, 50));
 console.log('- - - - - - - - - - - - - - - - - ');
 // 14. Write a loop that finds the second largest number in an array of numbers.
-const numbers14 = [10, 5, 20, 8, 15];
+const numbers14 = [10, 5, 20, 8, 15, 42];
+function orderingSmalToBig(numbers){
```
annasvst, line 155: I'd name this function `findSecontLargestNumber` as it doesn't order anything :)

annasvst: You solution works but there are other solutions that could be easier to read. I think the most common approach to solve this problem is the following:
- compare the first element of the array (array[0]) and the second (array[1]). Assign the biggest value to the variable named largestNumber, and the other value to the secondLargest.
- loop through the ret of the array (starting with array[2]) and compare each element to the largestNumber and secondLargest, if the new value is bigger than any of them, update largestNumber and secondLargest accordingly.

annasvst (changes requested):
HI Salih, well done overall! Please fix the issue in the ex. 13 and re-submit the assignment in classroom :)

PR #66, opened 2023-08-26
Saffet-Lesson24

Saffet01:


JS/Lesson24/saffet/filter.js
```diff
@@ -261,10 +270,26 @@ const productsExercise6 = [
 	},
 ];
 
+const outOfStock = productsExercise6.filter(product => {
+	if(product.quantity === 0){
```
annasvst, line 274: You need to keep only production that are in stock ( = "filter out the products that are out of stock") so that means products with quality more than 0
```suggestion
	if(product.quantity > 0){
```

JS/Lesson24/saffet/filter.js
```diff
@@ -261,10 +270,26 @@ const productsExercise6 = [
...
+});
+
+console.log(outOfStock);
+
 // Exercise 7:
 // Given an array of numbers, filter out the numbers that are divisible by 3.
 const numbersExercise7 = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
 
+const divisibleBy3 = numbersExercise7.filter(number => {
+	if(number%3 === 0){
```
annasvst, line 286:
```suggestion
	if(number%3 !== 0){
```

JS/Lesson24/saffet/filter.js
```diff
@@ -280,6 +305,14 @@ const wordsExercise8 = [
 	'jungle',
 ];
 
+const stringsHaveMore5Characters = wordsExercise8.filter(string => {
+	if(string.length>5){
```
annasvst, line 309:
```suggestion
	if(string.length<=5){
```

JS/Lesson24/saffet/filter.js
```diff
@@ -345,6 +378,14 @@ const studentsExercise9 = [
 	},
 ];
 
+const lowerThenC = studentsExercise9.filter(student => {
+	if(student.grade === "D+" || student.grade === "D"){
```
annasvst, line 382:
```suggestion
	if(student.grade !== "D+" || student.grade !== "D"){
```

It would be good to also add C-, D-, F+, F, F-

JS/Lesson24/saffet/filter.js
```diff
@@ -409,3 +450,11 @@ const usersExercise10 = [
 		email: 'sophia@example.com',
 	},
 ];
+
+const startWithJ = usersExercise10.filter(user => {
+	if(user.name.charAt(0)=== "J" || user.name.charAt(0)=== "j"){
```
annasvst, line 455: Use toLowerCase() to make sure that the text is in the same case, then you don't need to write multiple conditions
```suggestion
	if(user.name.charAt(0).toLowerCase() !== "j" ){
```

JS/Lesson24/saffet/map.js
```diff
@@ -179,10 +183,16 @@ const studentsExercise5 = [
 	},
 ];
 
+const studentAndGrade = studentsExercise5.map(student => {
```
annasvst, line 186: This does no actually extracts the array of values that you need but only prints the mentioned values one by one to console
```suggestion
const studentAndGrade = studentsExercise5.map(student => student.name + " " + student.grade);
console.log(studentAndGrade);
```

JS/Lesson24/saffet/map.js
```diff
@@ -329,6 +345,16 @@ const booksExercise9 = [
 	},
 ];
 
+const bookTitleAndAuthor = booksExercise9.map(book => {
+	console.log(`Title: ${book.title}
```
annasvst, line 349: Same here, you're just logging the values instead of extracting them into array

JS/Lesson24/saffet/filter.js
Saffet01: "out of stock" stokta olmayan ürünler olarak anladım :(

annasvst (approved):

PR #73, opened 2023-09-09
lesson 16 exercises finished

daglisinan:
Lesson 16 exercises finished

### DIFF
--- a/JS/Lesson16/sinan/script.js
+++ b/JS/Lesson16/sinan/script.js
@@ -1,142 +1,177 @@
 // 1. Write a loop that prints out the numbers from 1 to 10.
-for (let i = 1; i<=10; i++) {
+for (let i =0; i<10; i++) {
 	console.log(i);
-}
+	}
 console.log('- - - - - - - - - - - - - - - - - ');
 // 2. Write a loop that prints out the odd numbers from 1 to 20.
 // 1,3,5,7...
-for (let i = 1; i<20; i = i+2){
+for(let i=1; i<20; i+=2){
 	console.log(i);
 }
 
 console.log('- - - - - - - - - - - - - - - - - ');
 // 3. Write a loop that prints out the even numbers from 1 to 20.
 // 2, 4, 6...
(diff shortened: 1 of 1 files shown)


JS/Lesson16/sinan/script.js
```diff
@@ -1,142 +1,177 @@
 // 1. Write a loop that prints out the numbers from 1 to 10.
-for (let i = 1; i<=10; i++) {
+for (let i =0; i<10; i++) {
```
annasvst, line 2:
```suggestion
for (let i =1; i<11; i++) {
```
Otherwise numbers from 0 to 9 are printed!

JS/Lesson16/sinan/script.js
```diff
@@ -1,142 +1,177 @@
...
-for (let i = 1; i<20; i = i+2){
+for(let i=1; i<20; i+=2){
 	console.log(i);
 }
 
 console.log('- - - - - - - - - - - - - - - - - ');
 // 3. Write a loop that prints out the even numbers from 1 to 20.
 // 2, 4, 6...
-for (let i =2; i<=20; i+=2) {
+for(let i =0; i<20; i+=2) {
```
annasvst, line 15: 0 is not an even number, should be like this instead:
```suggestion
for(let i =2; i<=20; i+=2) {
```

JS/Lesson16/sinan/script.js
```diff
@@ -1,142 +1,177 @@
...
-console.log(newArray);
-
+console.log('Exercise 6.1:',newArray);
 console.log('- - - - - - - - - - - - - - - - - ');
 // 7. Write a while loop that counts from 10 to 1 and prints out each number.
-let i = 10;
-while(i>=1) {
-	console.log(i);
-	i--;
+let count = 0;
```
annasvst, line 59: It should be the other way around: from 10 to 1, the code that was here before was correct :)

annasvst (approved):
Approved!

PR #76, opened 2023-09-18
Onur  -- lesson24

Onuryemez54:


JS/Lesson24/onur/filter.js
```diff
@@ -344,6 +377,15 @@ const studentsExercise9 = [
 		grade: 'B-',
 	},
 ];
+const gradesMoreThanC = studentsExercise9.filter(students => {
+	const allGrades = ['A', 'A-', 'A+', 'B', 'B-', 'B+', 'C', 'C-', 'C+', 'D', 'D+', 'D-'];
+	return allGrades.indexOf(students.grade) <= allGrades.indexOf('C');
```
annasvst, line 382: Nice solution! :)

JS/Lesson24/onur/map.js
```diff
@@ -95,10 +99,13 @@ const products = [
 	},
 ];
 
-const names = products.map( product => {
+const names = products.map(product => {
```
annasvst, line 102: Since the return value is oneliner, you don't really need to use {}
```
const names = products.map(product => product.name);
```

JS/Lesson24/onur/map.js
```diff
@@ -114,6 +121,13 @@ const wordsExercise4 = [
 	'jungle',
 ];
 
+const findLengthOfEachWords = wordsExercise4.map(
```
annasvst, line 124: Same here, same line would be cleaner:
```
const findLengthOfEachWords = wordsExercise4.map(words => words.length);
```

annasvst (approved):
Hi Onur, great job! Feel free to use single line arrow function for cases when you have very simple logic (like in the map.js file), it makes your code cleaner and easier to read :)

JS/Lesson24/onur/filter.js
Onuryemez54: thank you :)

PR #79, opened 2023-09-24
Lesson26-Saffet

Saffet01:


JS/Lesson26/saffet/script.js
```diff
@@ -5,10 +5,99 @@ h2 (.post-title)
...
+        data.forEach(post => {
+
+            lastPostId = Math.max(...data.map(post => post.id));
+            console.log("Last Post id: ", lastPostId);
+            
+            const newPost = document.createElement("div");
+            newPost.classList.add("post");
+            newPost.setAttribute("data-post-id", post.id);
+
+            const editButton = document.createElement("button");
```
annasvst, line 31: Since the button allows the user to navigate to a different page rather than perform an action on the current page, it is more semantically correct to use <a> element instead of a <button> element here

JS/Lesson26/saffet/script.js
```diff
@@ -5,10 +5,99 @@ h2 (.post-title)
...
 };
 
+function removePost(postId){
+    const postToRemove = document.querySelector(`.post[data-post-id="${postId}"]`);
+    const editButtonToRemove = document.querySelector(`.button[data-post-id="${postId}"]`);
+    const deleteButtonToRemove = document.querySelector(`.delete-btn[data-post-id="${postId}"]`);
+    console.log("post selected succesfully");
+
+    if(postToRemove) {
+        postToRemove.remove();
```
annasvst, line 92: Very nice that you decided to handle the post removal from the page even though the API itself does not allow that :) 

It should be enough to do `postToRemove.remove();`, since edit and delete buttons are attached to the post element, they should be gone as well anyway

JS/Lesson26/saffet/script.js
```diff
@@ -5,10 +5,99 @@ h2 (.post-title)
...
+            postArea.appendChild(deleteButton);
+        });
+    })
+    .catch(error => console.log("Error: ",error));
 }
 
-function clearPosts() {
+window.addEventListener("load", fetchPosts);
+
+function clearPosts(postId) {
```
annasvst, line 66: I'd name this function `deletePost` as what it does is to remove a single post

JS/Lesson26/saffet/script.js
```diff
@@ -5,10 +5,99 @@ h2 (.post-title)
...
+        console.log("post succesfully deleted");
+        editButtonToRemove.remove();
+        console.log("Edit Button succesfully deleted");
+        deleteButtonToRemove.remove();
+        console.log("Delete Button successfully removed");
+    }
+}
+
+createButton.addEventListener("click", () => {
+    window.location.href = `createPostPage.html?id=${lastPostId+1}`;
```
annasvst, line 102: You don't need to pass an id to the create new post page. When we send a POST request from a Client to create a resource, an id of that resource is usually created on the server side. Leaving the logic for creating an id to the Client is a bad practice.

JS/Lesson26/saffet/createPostPage.html
```diff
@@ -0,0 +1,29 @@
...
+</head>
+
+<body>
+    <div class="container" id="container">
+        <h1>Posts</h1>
+
+        <form id="postForm" class="form">
+            <label for="title">Title:</label>
+            <input class="titleText" type="text" id="title" name="title">
+            <br><br>
```
annasvst, line 16: Try to avoid using `<br>`elements, they are leftovers from the past that are hard to maintain. Better add spacing using margin in CSS.

JS/Lesson26/saffet/createPostPage.html
```diff
@@ -0,0 +1,29 @@
...
+            <label for="body">Body:</label>
+            <textarea class="bodyText" id="body" name="body" rows="4" required></textarea><br>
+
+            <button class="button" id="saveButton">Save</button>
```
annasvst, line 20:
```suggestion
            <button type="submit" class="button" id="saveButton">Save</button>
```

JS/Lesson26/saffet/createPost.js
```diff
@@ -0,0 +1,46 @@
...
+const saveButton = document.getElementById("saveButton");
+const successMessage = document.getElementById("successMessage");
+const urlParams = new URLSearchParams(window.location.search);
+const postId = urlParams.get("id");
+
+saveButton.addEventListener("click", (event) => {
+
+    event.preventDefault();
+
+    if(titleInput.value.trim() === "" || bodyInput.value.trim() === ""){
```
annasvst, line 15: In HTML, title input does not have a required attribute, would be good to add it :)

JS/Lesson26/saffet/createPost.js
```diff
@@ -0,0 +1,46 @@
...
+        alert("Please fill out both title and body fields");
```
annasvst, line 16: Do you remember a lesson where we did a form validation? Try to add error messages to input fields in that way instead of showing an alert to the use.

JS/Lesson26/saffet/createPost.js
```diff
@@ -0,0 +1,46 @@
+const container = document.getElementById("container");
+const postArea = document.getElementById("postArea");
+const form = document.getElementById("postForm");
+const titleInput = document.getElementById("title");
+const bodyInput = document.getElementById("body");
+const saveButton = document.getElementById("saveButton");
+const successMessage = document.getElementById("successMessage");
+const urlParams = new URLSearchParams(window.location.search);
```
annasvst, line 8: You don't use post Id when sending the HTTP request so this and the next line are not needed :)

JS/Lesson26/saffet/createPost.js
```diff
@@ -0,0 +1,46 @@
...
+        }),
+        headers: {
+            "Content-type": "application/json; charset=UTF-8",
+        }
+    })
+        .then(response => response.json())
+        .then(newPost => {
+            console.log(newPost);
+
+            successMessage.style.display = "block";
```
annasvst, line 36: Better to use classes to show and hide HTML elements, if you add the .hidden class to CSS, you can add/remove it from the class list of the successMessage and the form here

JS/Lesson26/saffet/editPage.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>Edit Post</title>
+  <link rel="stylesheet" type="text/css" href="styles.css">
+</head>
+<body>
+  <div class="container" id="container">
+    <h1>Posts</h1>
```
annasvst, line 9:
```suggestion
    <h1>Edit post</h1>
```

JS/Lesson26/saffet/script.js
Saffet01: createButton.addEventListener("click", () => {
    window.location.href = `createPostPage.html`;
});

JS/Lesson26/saffet/script.js
Saffet01: Actually when I click the delete button post disappeared but buttons are still exist. That is why I removed buttons separately

annasvst (approved):
Approved!

PR #80, opened 2023-09-25
Lesson22-Exercise-Sinan

daglisinan:


JS/Lesson22/sinan/index.html
```diff
@@ -1,11 +1,51 @@
 <!DOCTYPE html>
-<html lang="en">
```
annasvst, line 2: The lang attribute in the HTML tag is used to specify the language of the document. It's important for accessibility and SEO (Search Engine Optimization) purposes. There're several reasons why it's needed:

- Accessibility: Screen readers and other assistive technologies use the lang attribute to determine the language of the content. This allows them to provide appropriate pronunciation and language-specific behaviors to assist users with disabilities. It ensures that users who rely on these technologies can access and understand the content correctly.

- SEO: Search engines also use the lang attribute to understand the language of a web page. This helps search engines present relevant search results to users searching in a particular language. It can improve the discoverability of your content in search results for users searching in that language.

- Internationalization: If you have a multilingual website, specifying the language using lang can help browsers and search engines understand which language variant of a page to display to users based on their language preferences or location.

JS/Lesson22/sinan/index.html
```diff
@@ -1,11 +1,51 @@
...
+<html>
 <head>
-  <meta charset="UTF-8">
-  <meta name="viewport" content="width=device-width, initial-scale=1.0">
-  <title>Accordion</title>
+  <meta charset='utf-8'>
```
annasvst, line 4: You need to use double quotes in HTML, not single quotes.

JS/Lesson22/sinan/index.html
```diff
@@ -1,11 +1,51 @@
...
+      <div class="label">List Four</div>
+      <div class="content">
+        <p>Lorem ipsum dolor sit amet consectetur adipisicing elit. 
+          Veniam laboriosam at tempora dicta deserunt perspiciatis
+           aliquam ea optio corrupti facere nostrum quis consequatur
+            repellendus quae dignissimos libero, non quia possimus!</p>           
+      </div>
+    </div>
+  </div>
+  <script src="main.js"></script></script>
```
annasvst, line 49:
```suggestion
  <script src="main.js"></script>
```

JS/Lesson22/sinan/style.css
```diff
@@ -0,0 +1,63 @@
+@import url('https://fonts.googleapis.com/css?family=Poppins:200,300,400,500,600,700,800,900&display=swap');
```
annasvst, line 1: Consider importing fonts in HTML instead of CSS. It will help you follow best practices and optimize the loading of your web page. Here are some reasons why you should consider importing fonts in HTML rather than CSS:

- Faster Rendering: Placing the font import directly in the HTML using the <link> or <style> tag in the <head> of your HTML document ensures that the browser loads the font early in the rendering process. This can lead to faster font rendering, reducing the chance of the "Flash of Unstyled Text" (FOUT) or "Flash of Invisible Text" (FOIT) that can occur when fonts are loaded asynchronously via CSS.

- Improved User Experience: By loading fonts early in the rendering process, you provide a better user experience. Users will see text rendered in the desired font sooner, even if other CSS styles are still being loaded.

- Preventing FOIT: Loading fonts in HTML can help prevent FOIT, where text remains invisible until the custom font is fully downloaded. This is especially important for critical text like headings or important content.

- Better Browser Compatibility: Some older browsers may not fully support @font-face rules in CSS or may have issues with how they're applied. Importing fonts in HTML is a more straightforward and reliable method that works across a wide range of browsers.

JS/Lesson22/sinan/style.css
```diff
@@ -0,0 +1,63 @@
...
+
+*
+{
+    margin: 0;
+    padding: 0;
+    box-sizing: border-box;
```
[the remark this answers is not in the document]
daglisinan: this code block already in * selector not any selected elements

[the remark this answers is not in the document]
annasvst: ok, I misread it as a body selector. Then, keep only `box-sizing: border-box;`, everything else should be moved to the body selector

JS/Lesson24/sinan/map.js
```diff
@@ -20,15 +20,15 @@ const fruits = wordsExercise1.map(word => {
 
 	return word.toUpperCase()
 });
-console.log(fruits);
+console.log('Exercise 1:',fruits);
```
annasvst, line 23: Changes in this file are not related to the Lesson22 homework, please revert (make sure that the code in map.js file is exactly as it is in the main branch in the same file right now).

JS/Lesson22/sinan/index.html
```diff
@@ -1,11 +1,51 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en"></html>
```
annasvst, line 2:
```suggestion
<html lang="en">
```
I guess the autocomplete closed the html tag for you by a mistake. Please fix :)

PR #81, opened 2023-09-26
Lesson17 Exercise Sinan

daglisinan:


JS/Lesson17/sinan/script.js
```diff
@@ -134,38 +134,84 @@ console.log('----------------------------');
 // Input: ({name: "John", age: 30}, "city", "New York")
 // Output: {name: "John", age: 30, city: "New York"}
 
+function getKeyValue (obj,key,value) {
+if(obj[key]=value) return obj;
```
annasvst, line 138: "=" symbol is used in JS for assignment, it should never be used in the condition statement like here. 
What you need to do in this exercise, is add the key value properties to the object:
```
  if (obj[key]) {
   return obj
} else {
  return obj[key] = value;
}
  
```

JS/Lesson17/sinan/script.js
```diff
@@ -134,38 +134,84 @@ console.log('----------------------------');
...
+  } 
+}
+  const ourList = removeKeyValue({name: "John", age: 30, city: "New York"}, "city");
+  console.log('Exercise 11:',ourList);
+
 console.log('----------------------------');
 // 12. Write a function that takes two arrays of equal length and returns an object with the first array as keys and the second array as values.
 // Input: (["name", "age", "city"], ["John", 30, "New York"])
 // Output: {name: "John", age: 30, city: "New York"}
+function concatKeyValue(list1,list2,) {
```
annasvst, line 162:
```suggestion
function concatKeyValue(list1,list2) {
```

JS/Lesson17/sinan/script.js
```diff
@@ -134,38 +134,84 @@ console.log('----------------------------');
...
+  const unifiedList = [];
```
annasvst, line 163: let newObject = {};

annasvst, line 163: What you need to do in this exercise is to create a new object from the two arrays:
```
if (list1.length === list2.length) {
  let newObject = {}; // Create a new empty object first
  for (let i = 0; i < list1.length; i++) {
    newObject[list1[i]] = list2[i];
  }
  return newObject;
} else  console.error('provided arrays are not of equal length!');

```

JS/Lesson17/sinan/script.js
```diff
@@ -134,38 +134,84 @@ console.log('----------------------------');
...
+  return unifiedList;
+}
+  const twoList  = concatKeyValue(["name", "age", "city"], ["John", 30, "New York"]);
+  console.log('Exercise 12:',twoList);
 
 // 13. Array Destructuring
 // Given the following array, use array destructuring to assign the values of x, y, and z to their respective variables: 
 const coordinates = [10, 20, 30];
+let chars = ['x','y','z'];
+function concatTwoList(list1,list2) {
```
annasvst, line 176: This is not what's asked in the exercise, what you need to do is the following:
`const [x, y, z] = coordinates;`. Also, you don't need `let chars = ['x','y','z'];`

PR #82, opened 2023-09-28
Tugba-Lesson18-JS

TugbaMartin:


JS/Lesson18/tugba/recursion.js
```diff
@@ -13,130 +13,316 @@
...
+	const result = removeSpecified(sentenceEx4, charToRemove);
+	console.log(result);
+	
+	console.log("================================")
+	// 6. Write a recursive function to check if an array includes a specific value.
+	console.log("Example6: ");
+	const numbers = [1, 2, 3, 4, 5];
+	const valueToCheck = 3;
+	
+	function removeSpecified(sentence, charToRemove){
```
annasvst, line 204: A function with name 'removeSpecified' is already declared in the same scope (ex. 5) so you cannot use it here.
Instead, the correct solution to Ex. 6 would be:
```
function includesValue(array, value) {
    // Base case: if the array is empty, the value is not found
    if (array.length === 0) {
        return false;
    }

    // Check if the first element is the value we're looking for
    if (array[0] === value) {
        return true;
    }

    // Recursive call on the rest of the array
    return includesValue(array.slice(1), value);
}

console.log(includesValue(numbers, valueToCheck));
```

annasvst (approved):

PR #83, opened 2023-09-28
Onder lesson18 js

ondermartin:


annasvst:
I'm closing this PR as it does not contain solution to lesson 18 but only has code for Lesson 14 and 15 that has already been merged. It's better to open a new branch for lesson 18, otherwise you'd have to resolve merge conflicts here.

PR #84, opened 2023-10-01
Tugba lesson24

TugbaMartin:


JS/Lesson18/tugba/recursion.js
```diff
@@ -4,7 +4,7 @@
 //   if (condition) {
 //     return value;
 //   } else {
-//     // If condition is not met, we modify parameter is some way
+//     // If condition is not met, we modify parameter is some way.....
```
annasvst, line 7: Hi Tuğba, this file got into the PR by mistake. Please copy the content of this file that is currently in the main branch and replace the content of this file in your PR to remove this difference. If you struggle with it, please watch the recording for our call together, there were a few cases like this that we resolved together :)

JS/Lesson24/tugba/filter.js
```diff
@@ -260,11 +263,16 @@ const productsExercise6 = [
...
 	},
 ];
+const inStockProducts = productsExercise6.filter(product => product.quantity > 0);
+
+console.log(inStockProducts);
 
 // Exercise 7:
 // Given an array of numbers, filter out the numbers that are divisible by 3.
 const numbersExercise7 = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
+const numbersDivisibleBy3 = numbersExercise7.filter(number => number % 3 === 0);
```
annasvst, line 273:
```suggestion
const numbersDivisibleBy3 = numbersExercise7.filter(number => number % 3 !== 0);
``` 
since the exercise says "filter out the numbers that are divisible by 3."

JS/Lesson24/tugba/filter.js
```diff
@@ -409,3 +424,6 @@ const usersExercise10 = [
 		email: 'sophia@example.com',
 	},
 ];
+const filteredUsers = usersExercise10.filter(user => user.name.startsWith('J'));
```
annasvst, line 427:
```suggestion
const filteredUsers = usersExercise10.filter(user => user.name.toUpperCase().startsWith('J'));
```
would be a safer choice in case the name is written starting with a lower case letter :)

JS/Lesson24/tugba/filter.js
```diff
@@ -344,6 +356,9 @@ const studentsExercise9 = [
 		grade: 'B-',
 	},
 ];
+const filteredStudents = studentsExercise9.filter(student => student.grade >= 'C');
```
annasvst, line 359: Unfortunately, comparing strings this way does not work in JS.

You can write a helper function that compares grades (check Lesson 15 assignment exercise 14) instead.

JS/Lesson24/tugba/map.js
```diff
@@ -178,12 +181,20 @@ const studentsExercise5 = [
...
+const studentNamesAndGrades = studentsExercise5.map(student => `${student.name} ${student.grade}`);
+
+console.log(studentNamesAndGrades);
 
 
 // Exercise 6:
 // Given an array of numbers, convert each number to a string representation using .map().
 const numbersExercise6 = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
 
+const stringRepresentation = numbersExercise6.map(number => String(number));
```
annasvst, line 193: In order to convert something to a string in JS, you do like this: `variableNameToConvert.toString();`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toString

```suggestion
const stringRepresentation = numbersExercise6.map(number =>number.toString());
```

annasvst (changes requested):
Hi Tuğba, there're a few minor things mentioned in the comments that need to be corrected, please take a look when you have time :)

annasvst (approved):
Approved

PR #89, opened 2023-10-07
Lesson15 Exercise Sinan

daglisinan:


JS/Lesson15/sinan/script.js
```diff
@@ -116,14 +129,49 @@ function checkIfStringIsUpperCase(stringToCheck) {
 	}
 }
 
-checkIfStringIsUpperCase('Write a program that takes in');
-checkIfStringIsUpperCase('write a program that takes in');
+checkIfStringIsUpperCase('Exercise 4.1:','Write a program that takes in');
```
annasvst, line 132: Since here you call the function directly and your function takes only one parameter (string value), you should not be adding 'Exercise 4.1:' to it. Instead, if you want it to be printed to the console, you can add it to the test string:
```suggestion
checkIfStringIsUpperCase('Exercise 4.1: Write a program that takes in');
```

JS/Lesson15/sinan/script.js
```diff
@@ -116,14 +129,49 @@ function checkIfStringIsUpperCase(stringToCheck) {
...
+checkIfStringIsUpperCase('Exercise 4.2:','write a program that takes in');
```
annasvst, line 133:
```suggestion
checkIfStringIsUpperCase('exercise 4.2: Write a program that takes in');
```

JS/Lesson15/sinan/script.js
```diff
@@ -175,5 +317,32 @@ whichMonthName(10);
 whichMonthName(2);
 
 // 14. Write a program that asks the user to enter a grade letter (A, B, C, D, or F) and displays a message indicating whether the grade is passing (A, B, C) or failing (D, F). Use a switch statement to handle the different cases.
+function gradeTest (point){
+	switch(point){
```
annasvst, line 321: It's a good practice to always make sure that the value you're using is in the correct format. YOu should do that inside the function instead of relying on the value to be passed in correct format from outside.
```suggestion
	switch(point.toUpperCase()){
```

JS/Lesson15/sinan/script.js
```diff
@@ -175,5 +317,32 @@ whichMonthName(10);
...
+			break;
+	}
+}
+gradeTest('a'.toUpperCase());
+
 
 // 15. Write a program that asks the user to enter their age, and then calculates and displays their age in dog years (where one human year is equivalent to seven dog years). If the user enters a negative number or zero, display an error message.
+
+function convertAgeToDogYears(humanYears) {
+	let humanAge = humanYears * 7;
```
annasvst, line 339: If you have a check inside of the function that validates the incoming value (e.g next line), make sure to do all the computations with the value after the check. Doing them before the check is either unnecessary or may result into errors due to incorrect format of the value.

annasvst (changes requested):
Hi Sinan, I've left a few minor comments&suggestions, please fix them and resubmit!

annasvst:
I see that the assignment has been already reviewed before in some other PR. Please make sure to close all the pull requests when assignment is completed!

PR #94, opened 2023-10-10
Onur-Lesson-26

Onuryemez54:


JS/Lesson26/onur/createNewPost/index.html
```diff
@@ -0,0 +1,33 @@
...
+    <div class="container">
+        <h1>Create a new post</h1>
+        <form id="new-post" method="post" novalidate>
+            <div class="form-container">
+                <label for="new-title">Title :</label>
+                <input type="text" name="title" id="new-title" class="text" required>
+            </div>
+            <div class="form-container">
+                <label for="new-body">Body :</label>
+                <textarea type="text" name="body" id="new-body" rows="7" class="text" required></textarea>
```
annasvst, line 19: Try avoid setting size of an input field directly in HTML by using "rows", use CSS instead. No need to change here.

JS/Lesson26/onur/createNewPost/index.html
```diff
@@ -0,0 +1,33 @@
...
+            </div>
+            <div class="form-container">
+                <input type="submit" value="Send" class="button button--success">
```
annasvst, line 22:
```suggestion
                <input type="submit" value="Create new post" class="button button--success">
```
Using a "Create new post" text on the button would give more background information about what action is about to happen to a user. Therefore, would be better for UX and accessibility.

JS/Lesson26/onur/createNewPost/styles.css
```diff
@@ -0,0 +1,70 @@
+* {
```
annasvst, line 1: A suggestion for your future projects: for generic CSS that are added for each page of your site/application, create a main.CSS file where you keep these styles. Then, just add that file to each HTML page. Only add a new CSS file that contain specific styles that are only applicable to this specific page.

JS/Lesson26/onur/createNewPost/styles.css
```diff
@@ -0,0 +1,70 @@
...
+
+.text:focus {
+    outline: 0;
+    border: 1px solid rgb(18, 215, 241);
+    box-shadow: 0 0 1px 1px #279fbd
+}
+
+h1 {
+    text-align: center;
+    color: #333;
```
annasvst, line 48: Suggestion for the future: in the beginning of the project, create variables for the colors that you will be using in the mentioned above main.css file and refer to variables everywhere where you need to use colors. It will make your code a lot more scalable.

JS/Lesson26/onur/editPost/edit.js
```diff
@@ -0,0 +1,35 @@
+
+
+const urlParams = new URLSearchParams(window.location.search);
+const titleParam = urlParams.get('title');
```
annasvst, line 4: Title and body of the post are a lot of data to share through query parameters. It's better to only share small pieces of data like id of the post to make urls easier to work with. When you have an id of the post, you can fetch the post data on the edit post page. Using 'onload' even listener in this case would be the right time to fetch the post data.

JS/Lesson26/onur/editPost/index.html
```diff
@@ -0,0 +1,33 @@
...
+            <div class="form-container">
+                <label for="edit-title">Title :</label>
+                <input type="text" name="title" id="edit-title" class="text" required>
+            </div>
+            <div class="form-container">
+                <label for="edit-body">Body :</label>
+                <textarea type="text" name="body" id="edit-body" rows="7" class="text" required></textarea>
+            </div>
+            <div class="form-container">
+                <input type="submit" value="Edit" class="button button--success">
```
annasvst, line 22: I'd say "Save changes" would be better accessibility-wise in this case :)

JS/Lesson26/onur/index.html
```diff
@@ -1,14 +1,23 @@
...
   <link rel="stylesheet" type="text/css" href="styles.css">
 </head>
+
 <body>
   <div class="container">
     <h1>Posts</h1>
+    <div class="button-container">
+      <button id="fetch-button" class="button button--success">Show All Posts</button>
+      <button id="clear-button" class="button button--danger">Clear All Posts</button>
+      <a href="./createNewPost/index.html" class="button button--newpost">Create A New Post</a>
```
annasvst, line 15: Nice that you choose to use an <a> element here! Well done :)

JS/Lesson26/onur/script.js
```diff
@@ -1,14 +1,68 @@
...
+
+            const postBody = document.createElement('p');
+            postBody.classList.add('post-body');
+            postBody.innerText = post.body;
+
+            listItem.appendChild(postTitle);
+            listItem.appendChild(postBody);
+
+            const editLink = document.createElement('a');
+            editLink.classList.add('button');
```
annasvst, line 32: I'd recommend to make the edit button an <a>  element, it should redirect the user to a new page.

JS/Lesson26/onur/script.js
```diff
@@ -1,14 +1,68 @@
...
+            editLink.classList.add('edit-button');
+            editLink.textContent = 'Edit';
+
+            listItem.appendChild(editLink);
+
+            editLink.addEventListener('click', editPost);
```
annasvst, line 38: I don't think you need an event listener for the edit button. Use <a> element and just add the content of the editPost function before this line: `listItem.appendChild(editLink);` so that the <a> element get's correct href attribute.

annasvst (changes requested):
Great job Onur, I've left minor comments, please take a look when you have time :)

annasvst (approved):
Approved! well done Onur :)

PR #99, opened 2023-10-21
SAVAS-Lesson13 Excercises were done and submitted.

savaspamuk:


JS/Lesson13/savas/script.js
```diff
@@ -0,0 +1,140 @@
...
+console.log('------------------------');
+// ---
+// 2. Declare a variable name and initialize it to your name as a string.
+// Declare a variable age and initialize it to your age as a number.
+// Declare a variable phrase and initialize it to the string "Hello, world!".
+// Add name and age to the sentence above. "Hello, world! My name is (name). I'm (age) years old.".
+let name = 'Savas';
+let age = 43;
+let phrase = 'Hello, world!';
+console.log("Exercise 2:", phrase, "My name is", name, ". I'm", age, "years old.");
```
annasvst, line 20: This code works when you print out the result to console like this. However, if you need to use a result in a form of a new sentense (like you'd usually do in real life) this would not work.

There're a few methods to combine strings in JS, please read this article: https://www.samanthaming.com/tidbits/15-4-ways-to-combine-strings/
or this one https://www.turing.com/kb/guide-to-string-concatenation-in-js

You'd normally use "+" operator or string interpolation (`` in combination with a ${variable) in a case like you have here.

annasvst (commented):
Hi Savaş, exercise 2 needs a correction, everything else apart from it looks good :)

annasvst (approved):
Looks good, approved!

PR #100, opened 2023-10-23
SAVAS-Exercise 14 was finished.

savaspamuk:


JS/Lesson14/savas/script.js
```diff
@@ -0,0 +1,101 @@
...
+const string2Exercise13 = "Bundle deal: Save $50 when you buy two products together!";
+console.log("Exercise 13.1: ", string1Exercise13.includes("%"));
+console.log("Exercise 13.2: ", string2Exercise13.includes("%"));
+
+
+console.log('------------------------');
+// 14. Use the `substring` method to extract the word "Databases" from string stringExercise14 based on two indices and print it to the console.
+// Question: What is the difference between `substring()` and `slice()`?
+const stringExercise14 = 'Databases are an essential component of many applications.';
+console.log("Exercise 14: ", stringExercise14.substring(0, 8));
```
annasvst, line 91: This is a correct solution but could you try to make it work for a different string? Let's say for the string: "The essential component of many applications are Databases".

Try combining substring with `indexOf()` method to find the start position and also use the length property of the word that you're searching ("Databases") to find the end index.

In real word, we usually have a dynamic input so we would have to find a way to use dynamic indexes as static values would not produce a reliable result.

JS/Lesson14/savas/script.js
```diff
@@ -0,0 +1,101 @@
...
+
+console.log('------------------------');
+// 15. Remove extra spaces from the string stringExercise15 and print the result to the console.
+const stringExercise15 = "    Hello world    ";
+console.log("Exercise 15: ", stringExercise15.trim());
+
+console.log('------------------------');
+// 16. Extract the price and currency from a string below using "slice" method and print it to console.
+const planPriceString = "Premium plan - 9.99 USD/month."; // Should return "9.99 USD"
+console.log("Exercise 16: ", planPriceString.slice(15, 23));
```
annasvst, line 101: Same here, try to use dynamic values :)

annasvst (commented):
Well done Savaş! I've added two comments about how to improve the solution. Try to solve it in a more advanced way :)

JS/Lesson14/savas/script.js
savaspamuk: I tried the following solution. I hope it's fine :)

JS/Lesson14/savas/script.js
savaspamuk: I tried a similar way here. Is it OK?

JS/Lesson14/savas/script.js
```diff
@@ -0,0 +1,104 @@
...
+console.log("Exercise 13.2: ", string2Exercise13.includes("%"));
+
+
+console.log('------------------------');
+// 14. Use the `substring` method to extract the word "Databases" from string stringExercise14 based on two indices and print it to the console.
+// Question: What is the difference between `substring()` and `slice()`?
+const stringExercise14 = 'Databases are an essential component of many applications.';
+console.log("Exercise 14a: ", stringExercise14.substring(0, 8));
+const stringExercise14b = 'The essential component of many applications are Databases';
+console.log("Exercise 14b: ", stringExercise14b.substring(stringExercise14b.indexOf("D"), stringExercise14b.length));
```
annasvst, line 93: Using `.indexOf("D")` is a bit dangerous because there could be other words starting with letter "D". Use the whole word instead:
```suggestion
console.log("Exercise 14b: ", stringExercise14b.substring(stringExercise14b.indexOf("Databases"), stringExercise14b.indexOf("Databases") + "Databases".length));

```

savaspamuk: Ok, I see the difference between them :)

annasvst (approved):
Approved!

PR #102, opened 2023-10-23
Onder-JS-lesson25

ondermartin:


JS/Lesson22/onder/app.jss
annasvst, line 1: The code doesn't work right now because the type of the file is 'jss' instead of 'js'.

JS/Lesson22/onder/app.jss
```diff
@@ -0,0 +1,13 @@
+let acc = document.getElementsByClassName("accordion");
+
+        for (let i = 0; i < acc.length; i++) {
+            acc[i].onclick = function () {
```
annasvst, line 4: The code here almost works. Right now, when a button clicked the first time, nothing happens due to the `if (panel.style.display === "block") {` check not getting the display property from the first attempt.
Since you also tried to make it work with css classes (a better approach then modifying css in js directly) I recommend to do like this:
app.js

```
let acc = document.getElementsByClassName("accordion");

for (let i = 0; i < acc.length; i++) {
    acc[i].onclick = function () {
        let panel = this.nextElementSibling;
        panel.classList.toggle("active");
    };
}
```

index.html
Add class `active` to **all panel divs** in HTML 
```
<div class="panel active">
```

CSS
```
.panel {
  display: none;
}

.panel.active { // make sure to select both panel and active classes to make sure that the css priority rules allows you to change the display property
  display: block;
}
```

JS/Lesson22/onder/index.html
```diff
@@ -1,11 +1,29 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
-  <meta charset="UTF-8">
-  <meta name="viewport" content="width=device-width, initial-scale=1.0">
-  <title>Accordion</title>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="stylesheet" href="./styles.css">
```
annasvst, line 6:
```suggestion
    <link rel="stylesheet" href="./style.css">
```
as your CSS file called 'style.css'

JS/Lesson22/onder/index.html
```diff
@@ -1,11 +1,29 @@
...
+    <title>accordion</title>
```
annasvst, line 7:
```suggestion
    <title>Accordion</title>
```

annasvst:
Closing this PR in favor of https://github.com/frontend-from-0/240123-exercises/pull/120/files

PR #103, opened 2023-10-25
SAVAS-Lesson 15 Exercise was done.

savaspamuk:


JS/Lesson15/savas/script.js
```diff
@@ -0,0 +1,437 @@
...
+checkAge(24);
+checkAge(76);
+checkAge(50);
+checkAge(8);
+checkAge(-3);
+
+// 10. Write a program that takes age as a number and whether they are a student (yes or no). If the age is less than 18 and the user is a student, print "You are a student and not old enough to vote". If the age is between 18 and 65 (inclusive) and the user is not a student, print "You are eligible to vote". If the age is greater than 65, print "You are a senior citizen". If the user is a student but their age is greater than or equal to 18, print "You are old enough to vote".
+console.log("---------------------- Excersise 10: ---------------------")
+
+function appropriateForVote(age, student) {
```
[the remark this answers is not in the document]
annasvst: try to re-write the function accordingly :)

JS/Lesson15/savas/script.js
```diff
@@ -0,0 +1,437 @@
...
+    else if (salary >= 50000 || experience >= 10) {
+        console.log("We can offer you the job at the requested salary.")
+    }
+
+    else {
+        console.log("Invalid value.")
+    }
+}
+
+offerSalary(20000, "3");
```
annasvst, line 323: Here's a problem! Inside a function, experience parameter is treated as a number. However, when calling a function, a string is passed to it (e.g. `"3"`). 
Since there's no type restriction in JS, it's good to always try to use variables in the expected  types when you can and add checks to see if the value is in the expected format.
E.g. I'd modify the function like this:
```
function offerSalary(salary, experience) {
  if (typeof salary !== 'number' || typeof experience !== 'number') { // should be the very first check
    console.log("Invalid input. Both salary and experience must be numbers.");
  } else if ...
....
```

JS/Lesson15/savas/script.js
```diff
@@ -0,0 +1,437 @@
...
+}
+
+whichMonthName(5);
+whichMonthName(10);
+whichMonthName(2);
+
+// 14. Write a program that asks the user to enter a grade letter (A, B, C, D, or F) and displays a message indicating whether the grade is passing (A, B, C) or failing (D, F). Use a switch statement to handle the different cases.
+console.log("---------------------- Excersise 14: ---------------------")
+
+function result(gradeLetter) {
```
annasvst, line 399: This is a good solution. Could you try to make it work even if the value that is passed to the function is not upper case (e.g. `result("a");`)?

annasvst (changes requested):
Good job Savaş, I've left a few comment for minor improvements, please check and resubmit when done :)

JS/Lesson15/savas/script.js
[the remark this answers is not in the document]
savaspamuk: Ok, I did it. It's very clear now :)

JS/Lesson15/savas/script.js
savaspamuk: Ok, I see. Also, I left the else with invalid input, because negative numbers are also invalid input. Is it true method?

JS/Lesson15/savas/script.js
savaspamuk: Ok, I did it by using
        
        gradeLetter = gradeLetter.toUpperCase()

And now it works :)

JS/Lesson15/savas/script.js
```diff
@@ -0,0 +1,437 @@
...
+// 8. Write a program that takes in a person's height and weight and checks if they are considered underweight, normal, overweight, or obese using if else statement.
+// Body Mass Index (BMI) is a person’s weight in kilograms (or pounds) divided by the square of height in meters (or feet). A high BMI can indicate high body fatness.
+//Formula: weight (kg) / [height (m)]2
+// Below 18.5	Underweight
+// 18.5 – 24.9	Healthy Weight
+// 25.0 – 29.9	Overweight
+// 30.0 and Above	Obesity
+console.log("---------------------- Excersise 8: ----------------------")
+
+function testBMI(height, weight) {
```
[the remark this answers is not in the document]
savaspamuk: I actually thought about doing that, but I didn't. I defined a variable based on your suggestions and also cleaned up some parentheses in the code. Function codes are clearer than before.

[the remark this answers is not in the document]
annasvst: Looks good now :)

JS/Lesson15/savas/script.js
annasvst: This solution works however to write cleaner code, it would be better to assign the `gradeLetter.toUpperCase()` to a new variable, then use that variable later in the code. Another approach would be to convert the `gradeLetter` to upper case directly in the switch statement:
```
switch (gradeLetter.toUpperCase()) {
```

annasvst: It's not critical to change the code here but I want you to know that for the future :)

annasvst (approved):
Approved!

PR #105, opened 2023-10-29
Saffet-Lesson30

Saffet01:


JS/Lesson30/saffet/my-app/src/components/Ex9/index.jsx
```diff
@@ -0,0 +1,24 @@
...
+    setBackgroundColor("yellow");
+  }
+
+  function changeBackgroundtoRed(){
+    setBackgroundColor("red");
+  }
+
+  return (
+    <>
+      <div className="btn-div" style={{backgroundColor: backgroundColor}}>
```
annasvst, line 18: It's always better to use a class that inline styling (`style={{backgroundColor: backgroundColor}}`) :)

annasvst (approved):
well done Saffet! The comments I left are for future, no need to correct here but pay attention to the mentioned things in the future :)

PR #106, opened 2023-10-29
Onder-JS-Lesson26

ondermartin:


JS/Lesson26/onder/index.html
```diff
@@ -1,14 +1,20 @@
...
+<html lang="en">
 <head>
-  <title>Posts</title>
-  <link rel="stylesheet" type="text/css" href="styles.css">
+    <meta charset="UTF-8">
+    <title>Mocked Posts</title>  
+    <link rel="stylesheet" type="text/css" href="./styles.css">
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
+    <a href="./createPost/create.html">Create Post</a>
+    <a href="./EditPost/edit.html">Edit Post</a>
```
annasvst, line 9: This url should not be on the main page, it should only be available for each individual post so that we can get a post id

JS/Lesson26/onder/index.html
```diff
@@ -1,14 +1,20 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
 <head>
-  <title>Posts</title>
-  <link rel="stylesheet" type="text/css" href="styles.css">
+    <meta charset="UTF-8">
+    <title>Mocked Posts</title>  
+    <link rel="stylesheet" type="text/css" href="./styles.css">
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
```
annasvst, line 7: Please don't use jquery, it's abandoned nowadays.

JS/Lesson26/onder/script.js
```diff
@@ -1,14 +1,27 @@
...
-Elements to use:
-li (.post)
-h2 (.post-title)
-p (.post-body)
-*/
-
-const URL = 'https://jsonplaceholder.typicode.com/posts';
-function fetchPosts() {
+function editPost(id) {
+    window.location.href = `edit.html?id=${id}`;
```
annasvst, line 2: The url isn't correct, should be `EditPost/edit.html?id=${id}` according to the current project structure

annasvst (approved):
Hi Önder, this assignment is approved however there're some recommendations:
- do not use JQuery going forward. It's currently an abandone technology so I would not recommend learning it today.
- try to keep folder names consistent (e.g. if one is `createPost` than `EditPost` should also be in the same case - `editPost`)
- avoid using <br> elements in HTML, instead use margin (and padding) in CSS to add spacing
- take another look into Block and Inline level elements in HTML/CSS, I think it could be that you forgot it a bit and have a hard time aligning elements on the screen: https://www.w3schools.com/html/html_blocks.asp. E.g. Input elements are inline level elements, that's what make the next input filed to be displayed on the same line. Maybe also take another look at CSS flexbox and grid. Doing alignment using them is even easier!


Great job otherwise, I'm happy to see this project!

PR #107, opened 2023-10-30
SAVAS-Lesson16 was finished.

savaspamuk:


JS/Lesson16/savas/script.js
```diff
@@ -1,75 +1,173 @@
...
 	console.log(i);
 }
 
-console.log('- - - - - - - - - - - - - - - - - ');
+console.log('- - - - - - - - - - - - - - - - - EXERCISE 4 - - - - - - - - - - - - - - - - - ');
 // 4. Write a loop that finds the maximum number in an array of numbers.
 const findMaxNumberInArray = [300, 10, 23, 42, 505, 12, 7, 4, 900];
+let num = findMaxNumberInArray[0];
+for (let i = 0; i < findMaxNumberInArray.length; i++) {
+	if (num < findMaxNumberInArray[i+1]) {
```
annasvst, line 26: When the loop will come to the last element of the array, expression `findMaxNumberInArray[i+1]` will try to get an element outside array (non-existent). It does not cause a problem here but may do that in a different situation.
A safer approach would be to set the initial value of i to 1 (`for (let i = 1; i < findMaxNumberInArray.length; i++) {`) as you covered the element with index 0 already. Then, you'd need to check current index instead: `if (num < findMaxNumberInArray[i]) {`

JS/Lesson16/savas/script.js
```diff
@@ -1,75 +1,173 @@
...
+console.log(`The sum of all even numbers between 1 and 50 is ${sum2}.`);
 
-console.log('- - - - - - - - - - - - - - - - - ');
+console.log('- - - - - - - - - - - - - - - - - EXERCISE 14 - - - - - - - - - - - - - - - - ');
 // 14. Write a loop that finds the second largest number in an array of numbers.
-const numbers14 = [10, 5, 20, 8, 15];
-
-console.log('- - - - - - - - - - - - - - - - - ');
+let numbers14 = [10, 5, 20, 8, 15];
+numbers14.sort(function(a, b) {
```
annasvst, line 141: Smart solution! It's not really part of the "iterations" topic but it is something that's used more often in real word.

One thing that you need to be careful with is that there could be more than one of the same number in the array. E.g. for array like this: `[20, 5, 20, 8, 15];` the code doesn't work any more.

JS/Lesson16/savas/script.js
```diff
@@ -1,75 +1,173 @@
...
+console.log(`The second largest number is ${numbers14[1]}.`);
+
+console.log('- - - - - - - - - - - - - - - - - EXERCISE 15 - - - - - - - - - - - - - - - - ');
 // 15. Write a loop that counts the number of vowels in a given string.
 const string15 = 'Hello, how are you?';
 const vowels = 'aeiouAEIOU';
+let countsNumVovels = 0;
+for (let i = 0; i < string15.length; i++) {
+	for (let ii = 0; ii < vowels.length; ii++) {
+		if ( string15[i].includes(vowels[ii])) {
```
annasvst, line 155: Actually, `vowels.includes(string15[i])` would check if a given character is in vowels so you don't really need the second loop here.

annasvst (changes requested):
Hi Savaş, I left a few comment. Great job overall!

JS/Lesson16/savas/script.js
savaspamuk: I removed the second loop, but I couldn't work it :/

annasvst: The solution is as short as this:
```
for (let i = 0; i < string15.length; i++) {
	if (vowels.includes(string15[i])) countsNumVovels++;
}
console.log(`The number of vowels in "${string15}" is ${countsNumVovels}.`);
```

annasvst (approved):
Approved!

PR #109, opened 2023-11-04
Tugba JS-LESSON 30

TugbaMartin:


JS/Lesson30/tugba/my-app/src/components/Ex8/index.js
```diff
@@ -0,0 +1,35 @@
...
+    return setList(updatedList);
+  };
+
+  return (
+    <>
+      <ul>
+        {list.map((item) => (
+          <li key={item.id}>
+            <span>{item.label}</span>
+            <button onClick={() => deleteButton(item.id)}>Click to Delete</button>
```
annasvst, line 29: Since the function that is called inside onClick event peforms an action, it's comment to use verbs when naming it: e.g. 'handleDeleteClick' would be appropriate name in this case.

JS/Lesson30/tugba/my-app/src/components/Ex10/index.js
```diff
@@ -0,0 +1,22 @@
+import {useState} from 'react';
+
+export const Ex10 = () => {
+  // useState hook
+  const [clickCount, setClickCount] = useState(0);
+
+  function handleClick () {
```
annasvst, line 7: `event.detail` property would give you the information about how many times the button was clicked.

so the fucntion would be modified like this:
```
function handleClick (event) {
  if (event.detail === 2) alert("double-clicked");
}
```

annasvst (changes requested):
Hi Tuğba, looks good overall. Ex. 10 needs to be fixed and ex 8 could be improved. See comments in the review :)

JS/Lesson30/tugba/my-app/src/components/Ex8/index.js
```diff
@@ -0,0 +1,37 @@
...
+
+  const deleteButton = (itemId) => {
+    const updatedList = list.filter((item) => item.id !== itemId);
+    console.log(itemId);
+    console.log(updatedList);
+    return setList(updatedList);
+  };
+
+  return (
+    <>
```
annasvst, line 24: You only need to use Fragment component  (`<></>`) in React if you have more than one element on the same level that you need to return. Here you only have `<ul>` so no need to wrap it in `<></>`.

annasvst (approved):
Approved

PR #111, opened 2023-11-04
Saffet-Lesson31

Saffet01:


JS/Lesson31/saffet/my-app/src/components/CreateTask/index.jsx
```diff
@@ -0,0 +1,29 @@
...
+    onCreateProps(task);
+    setTask("");
+  }
+
+  return (
+    <>
+      <h2>Create a task in below</h2>
+      <form className="form" action="">
+        <input onChange={handleInputChange} className="form-input" type="text" />
+        <button onClick={handleSubmit} className="create-btn">Save</button>
```
annasvst, line 25: If you add onSubmit event handeler to the form element itself, you won't need to any event handler to this button. Just make sure to set the type of this button to "submit".
```suggestion
        <button type="submit" className="create-btn">Save</button>
```

JS/Lesson31/saffet/my-app/src/components/CreateTask/index.jsx
```diff
@@ -0,0 +1,29 @@
...
+    console.log(task);
+
+    onCreateProps(task);
+    setTask("");
+  }
+
+  return (
+    <>
+      <h2>Create a task in below</h2>
+      <form className="form" action="">
```
annasvst, line 23: We usually don't specify "action" attributes when using form elements in React.
```suggestion
      <form className="form" onSubmit={handleSubmit}>
```

JS/Lesson31/saffet/my-app/src/components/CreateTask/index.jsx
```diff
@@ -0,0 +1,29 @@
...
+        <input onChange={handleInputChange} className="form-input" type="text" />
```
annasvst, line 24: I'd recommend to change the input field like this:
```suggestion
       <input value={task} onChange={handleInputChange} className="form-input" type="text" />
```
I passed the `task` to input as `value` so that it can be reset easily once form is submitted. OnChange, the `task` value is updated.

JS/Lesson31/saffet/my-app/src/components/CreateTask/index.jsx
```diff
@@ -0,0 +1,29 @@
...
+  const [task, setTask] = useState("");
+
+  const handleInputChange = (event) => {
+    setTask(event.target.value);
+  }
+
+  const handleSubmit = (event) => {
+    event.preventDefault();
+    console.log(task);
+
```
annasvst, line 15: I'd modify this function a bit so we only add a new todo item if there's any text:
```
if (task.length > 0) {
      onCreateProps(task);
      setTask("");
    }
```

JS/Lesson31/saffet/my-app/src/components/List/index.jsx
```diff
@@ -1,31 +1,48 @@
...
-          listItem => <ListItem handleDelete={handleDelete} key={listItem.id} id={listItem.id} title={listItem.title} />
-        )
-      }
-  </ul>
-)
-};
\ No newline at end of file
+  return (
+    <>
+      <CreateTask onCreateProps={addTask} />{" "}
```
annasvst, line 35: I'd call the prop that component takes something like 'handleCreateTask' or 'handleCreateTodo' since the purpose is to create todo items. It's always better to keep the language consistent.
```suggestion
      <CreateTask handleCreateTodo={addTask} />{" "}
```
And if you change prop name here it's important to update the `CreateTask` component.

JS/Lesson31/saffet/my-app/src/components/ListItem/index.jsx
```diff
@@ -1,8 +1,18 @@
...
+
+	const handleClick = () => {
+		setCompleted(!completed);
+	}
+
+	const className = completed ? 'completed' : '' ;
+
 	return (
-		<li>
+		<li onClick={handleClick} className={className}>
```
annasvst, line 15: If you add a span around `title` and move the `className` to it, then only the text of the list item would be crosses when an item is checked :) 
```suggestion
		<li onClick={handleClick}>
		  <span className={className}>{title}</span>

```

annasvst (commented):
Hi Saffet, well done! I've left comments that would help fixing minor issues :)

PR #115, opened 2023-11-06
Onur-Lesson31

Onuryemez54:


JS/Lesson31/onur/my-app/src/components/List/index.jsx
```diff
@@ -1,31 +1,49 @@
...
+    { id: 3, title: "Go running", completed: false },
+    { id: 4, title: "Do dishes 2", completed: false },
+    { id: 5, title: "Do homework 3", completed: false },
+    { id: 6, title: "Go running 4", completed: false },
   ];
 
   const [todoItems, setTodoItems] = useState(data);
 
+  const handleCreateTodo = (newItem) => {
+    const newTodoId = (todoItems.length) + 1;
```
annasvst, line 19: There's a problem here. Let's say there're 5 items in the list, ids are 1,2,3,4,5.
Then, I remove item 4 (item with id 4 is gone) and add a new item. The id of the new item will be 5, so the list would look like this: 1,2,3,5,5. 
We want id to be a unique value. Normally, you'd use a library to generate a uuid (Universal Unique Identifier) but you can use a hack here:
 ```const newTodoId = Math.max(...todoItems.map((item) => item.id)) + 1;```

JS/Lesson31/onur/my-app/src/components/ListItem/index.jsx
```diff
@@ -1,9 +1,18 @@
+import { useState } from 'react';
 import './styles.css';
 
 export const ListItem = ({ id, title, handleDelete }) => {
+
+	const [completed, setCompleted] = useState(false);
+
+	const className = (!completed) ? '' : 'completed';
```
annasvst, line 8: It's probably easier for you to read the code right now but it's a very uncommon to use parenthesis around a single condition. Instead, try doing like this:
```suggestion
	const className = !completed ? '' : 'completed';
```

Of course, if you have a more complex condition that it makes sense to add parenthesis.

annasvst: Also, negation adds cognitive complexity to the code, try to avoid it as much as possible.
Final suggestion:
```suggestion
	const className = completed ? 'completed' : '';
```

JS/Lesson31/onur/my-app/src/components/ListItem/index.jsx
```diff
@@ -1,9 +1,18 @@
...
+
+	const changeClassName = () => {
+		return (className === '') ? setCompleted(true) : setCompleted(false);
+	}
 	return (
-		<li>
-			{title}
+		<li onClick={changeClassName}>
```
annasvst, line 14: I'd solve this like this instead (an I think it's the most ofter used pattern in React):
1. 
```suggestion
		<li onClick={() => setCompleted(prev => !prev)}>
```
2. You can now remove `changeClassName` function 
3. Update the span below as following:
`<span className={completed ? 'completed' : ''}>{title}</span>`
4. You can now also remove this line above: `const className = (!completed) ? '' : 'completed';`

annasvst (commented):
Hi Onur, great job! I've added comments with some possible improvement in the review :)

annasvst (approved):

PR #116, opened 2023-11-08
Saffet-Lesson32

Saffet01:


JS/Lesson32/saffet/my-app/src/components/RecipeDetail/index.jsx
```diff
@@ -0,0 +1,15 @@
+export const RecipeDetail = ({recipe}) => {
+
+    console.log(recipe);
```
annasvst, line 3: You don't need to render component at all if no recipe data is passed to it. Instead, add a check in the RecipeList (see my comment there).

JS/Lesson32/saffet/my-app/src/components/RecipeList/index.jsx
```diff
@@ -1,12 +1,26 @@
...
+      <ul>
+        {recipes.map((recipe) => {
+          return (
+            <li key={recipe.id} onClick={() => handleClick(recipe)}>
+              {recipe.strMeal}
+            </li>
+          );
+        })}
+      </ul>
+      <RecipeDetail recipe={selectedRecipe} />
```
annasvst, line 23:
```suggestion
      {selectedRecipe ? <RecipeDetail recipe={selectedRecipe} /> : null }
```

annasvst (changes requested):
Hi Saffet, I've added a comment with suggestion how to improved the code. Overall, I'd expect more from thsi assignment. E.g. add a meal image and try to make the data displayed on the page look a bit better like if it were a real application.

PR #120, opened 2023-11-13
Onder-JSLesson22-corrected

ondermartin:


JS/Lesson22/onder/script.js
```diff
@@ -0,0 +1,8 @@
+let acc = document.getElementsByClassName("accordion");
```
annasvst, line 1: Since the value of this variable is never change it's better to declare it using `const` keyword:

```suggestion
const acc = document.getElementsByClassName("accordion");
```

JS/Lesson22/onder/script.js
```diff
@@ -0,0 +1,8 @@
...
+
+for (let i = 0; i < acc.length; i++) {
+    acc[i].onclick = function () {
+        let panel = this.nextElementSibling;
```
annasvst, line 5: Since the value of this variable is never change it's better to declare it using `const` keyword:
```suggestion
        const panel = this.nextElementSibling;
```

PR #127, opened 2023-11-26
lesson32 done

TugbaMartin:


annasvst (changes requested):
Hi Tuğba, this is a very good start! Now RecipeDetail component is missing in your code. I see that there's a file in components folder created for it but it's empty. You need to add code to present recipe details. Some elements like div, heading, a couple of paragraphs and image is enough to display recipe data. Also, add back button in the Recipe details component and pass the 'handleClearSelection' to it to return to the main page.

TugbaMartin:
Hello Anna I just Updated recipe Details PR

PR #129, opened 2023-11-28
SAVAS-Lesson-24 was finished.

savaspamuk:


JS/Lesson24/savas/filter.js
```diff
@@ -327,6 +348,9 @@ const studentsExercise9 = [
 	},
 ];
 
+const successfulStudents = studentsExercise9.filter(student => student.grade == "A" || student.grade == "A+" || student.grade == "A-" || student.grade == "B" || student.grade == "B+" || student.grade == "B-" || student.grade == "C" || student.grade == "C+" || student.grade == "C-");
```
annasvst, line 351: This code certainly works but could you think about a way to optimize it as there's a lot of copy pasting right now. Try to find the pattern: e.g. "A", "A+", "A--", "A-". They all start with the letter "A", right? How could you use that to minimize the code?

annasvst (commented):
Great job Savaş, I left one comment, please try to improve the code according to the suggestion :)

JS/Lesson24/savas/filter.js
savaspamuk: Wow, I think I found a way to do it :)

PR #131, opened 2023-11-29
Onur-Lesson34

Onuryemez54:


JS/Lesson34/onur/my-app/src/components/NewRecipe/index.jsx
```diff
@@ -1,5 +1,6 @@
 import { useForm } from 'react-hook-form';
 import './styles.css';
+import { useState } from 'react';
```
annasvst, line 3: useState is unused, feel free to remove this import :)

JS/Lesson34/onur/my-app/src/components/NewRecipe/index.jsx
```diff
@@ -22,56 +23,195 @@ export const NewRecipe = () => {
...
 	const strCategory = watch('strCategory');
 
-	const onSubmit = (data) => console.log(data);
+
+
+
+	const onSubmit = (data) => {
+
+		if (data.strDrinkAlternate.trim() === '') {
+			data.strDrinkAlternate = null;
```
annasvst, line 32: I see that you're trying to modify the data that is coming from the form, if the value of input is empty (or spaces).  React-hook-form does not actually allow you to modify the incoming form data like this so this operation does not make any sence. 

Instead, you could add a pattern match for required input fields: /\S+/ regex should match any character exclusive white spaces so that your input will only be valid if there is an actual word or other character added.

JS/Lesson34/onur/my-app/src/components/NewRecipe/index.jsx
```diff
@@ -22,56 +23,195 @@ export const NewRecipe = () => {
...
+		ingredientsArray.map((ingredient, index) => {
+			for (let i = 1; i <= 20; i++) {
+				if (ingredient.trim().length > 0 && (i <= ingredientsArray.length)) {
+					data[`strIngredient${index + 1}`] = ingredient;
+				} else {
+					data[`strIngredient${i}`] = "";
+				}
+			}
+		});
+		delete data.ingredients;
```
annasvst, line 58: I suppose you want to clear input filed by doing `delete data.ingredients;`. It does not work like this. Instead, there's a `reset` function: [reset](https://www.react-hook-form.com/api/useform/reset/).
For reset to work, you must provide default values to `useForm();`.
E.g. `const {reset} = useForm({strMeal: '', strDrinkAlternate:'', ...});

JS/Lesson34/onur/my-app/src/components/NewRecipe/index.jsx
```diff
@@ -22,56 +23,195 @@ export const NewRecipe = () => {
...
+
+		const measuresArray = data.measures.split(",").map((ingredient) => ingredient.trim());
+		measuresArray.map((measure, index) => {
+			for (let i = 1; i <= 20; i++) {
+				if (measure.trim().length > 0 && (i <= measuresArray.length)) {
+					data[`strMeasure${index + 1}`] = measure;
```
annasvst, line 64: Don't modify the incoming form data. Instead, create a new variable that will hold the form data that you later will send to the service or use in some other way in your application then add the computed values to that new variable.

JS/Lesson34/onur/my-app/src/components/NewRecipe/index.jsx
```diff
@@ -22,56 +23,195 @@ export const NewRecipe = () => {
...
+				)}
+
+				<label className='form-label' htmlFor="strImageSource">Image source :</label>
+				<input className='form-input' type="text" id='strImageSource' {...register("strImageSource", { pattern: /^https:\/\// })} />
+
+				<label className='form-label' htmlFor="strCreativeCommonsConfirmed">Creative Commons :</label>
+				<input className='form-input' type="text" id='strCreativeCommonsConfirmed' {...register("strCreativeCommonsConfirmed")} />
+
+				<label className='form-label' htmlFor="dateModified">Date Modified :</label>
+				<input className='form-input' type="text" id='dateModified' {...register("dateModified")} />
```
annasvst, line 211: Values like date modified, date created, etc. are usually created programatically. So when the form is submitted, create a new data and assign it to your own variable for the form data before sending the data to server or using it in some other way.

annasvst: And remove the input element for the date :)

JS/Lesson34/onur/my-app/src/components/NewRecipe/index.jsx
```diff
@@ -22,56 +23,195 @@ export const NewRecipe = () => {
...
+				<input type="text" id='strYoutube' className='form-input' {...register("strYoutube", {
+					required: true,
+					pattern: /^(https?:\/\/)?(www\.)?(youtube\.com\/(?:[^\/\n\s]+\/\S+\/|(?:v|e(?:mbed)?)\/|\S*?[?&]v=)|youtu\.be\/)([a-zA-Z0-9_-]{11})/
+				})} />
+				{errors.strYoutube && (
+					<span className='input-error'>Please enter a valid Youtube URL</span>
+				)}
+
+				<label className='form-label'>Ingredients : </label>
+				<textarea className='form-input' placeholder='Please separate by commos' {...register("ingredients", { required: true, minLength: 1 })} />
```
annasvst, line 184:
```suggestion
				<textarea className='form-input' placeholder='Please separate by commas' {...register("ingredients", { required: true, minLength: 1 })} />
```

JS/Lesson34/onur/my-app/src/components/NewRecipe/index.jsx
```diff
@@ -22,56 +23,195 @@ export const NewRecipe = () => {
...
+				{errors.ingredients && (
+					<span className='input-error'>Ingredients are required</span>
+				)}
+
+				<label className='form-label'>Measures : </label>
+				<textarea className='form-input' placeholder='Please separate by commos' {...register("measures", { required: true, minLength: 1 })} />
```
annasvst, line 190:
```suggestion
				<textarea className='form-input' placeholder='Please separate by commas' {...register("measures", { required: true, minLength: 1 })} />
```

annasvst (changes requested):
Hi Onur, good try! I've added a few comments with suggestions, please take a look and try to fix them. Let me know if something is unclear and I'll help :)

JS/Lesson34/onur/my-app/src/components/NewRecipe/index.jsx
Onuryemez54: Actually, I do not want to clean the entry filed here. After creating values such as Ingredients1, Ingredients2, Ingredients3, a separate string named ingredients was also created in the data. I wanted to have only Ingredients{i} values, so I removed the ingredients string.

annasvst (approved):
Looking good, approved! :)

PR #132, opened 2023-11-30
completed

ondermartin:


annasvst (commented):
Hi Önder, this is a good step forward but the pull request requires a bit more work :) 
If you look at the structure of this file `JS/Lesson34/onder/my-app/src/data.json`, you'll see other fields that need to be provided by the user. E.g. strMealThumb, strTags, strYoutube, strMeasure1 - strMeasure20, strSource, strImageSource, strCreativeCommonsConfirmed. 

Make sure that these fields are not required so the user is not forced to enter value in there. 
Same for the ingredients, a user should ot be force to enter all 20 ingredients. 

Also, when you have multiple similar values like for ingredients here, you don't want to have it as 20 separate input fields in the form. Normally, we'd handle it with a more complex solution where it would be possible to add new input fields if needed for ingredients and measures. However, not to complicate it now you can have a `textarea` input and ask a user to input all ingredients separated by a coma (e.g. "milk, eggs, water, ..."). Then, when you consume the data from the form, you'd need to split the value of `textarea` and add assign it to different `strIngredient` fields.

ondermartin (commented):
Hello Anna 
I tried to do my best on this hopefully this is good

JS/Lesson34/onder/my-app/src/components/NewRecipe/index.jsx
```diff
@@ -1,77 +1,139 @@
...
+  const [formData, setFormData] = useState(recipeData);
+
+  const strCategory = watch('strCategory');
+
+  Object.keys(formData).forEach((key) => {
+    setValue(key, formData[key]);
+  });
+
+  const onSubmit = (data) => {
+    Object.keys(data).forEach((key) => {
```
annasvst, line 22: After looking at the entire code of this component, I'd recommend to remove the state (`const [formData, setFormData] = useState({});`) completely as it's not used anywhere. You only try to update it but never use it.

Normally, we'd either send an API request here or update application storage but we'll do it in the following lessons.

Since you use react-hook-form, you get `data` parameter in the `onSubmit` function that contain all the data that you need. 
What you need to do before using this data is to trim it as it may have spaces (e.g. if someone types '    Pizza   ' instead of 'Pizza').

JS/Lesson34/onder/my-app/src/components/NewRecipe/index.jsx
```diff
@@ -1,77 +1,139 @@
...
+    formState: { errors },
+  } = useForm();
 
-//   return [ myData, deleteMyData ];
-// }
+  const [formData, setFormData] = useState(recipeData);
+
+  const strCategory = watch('strCategory');
+
+  Object.keys(formData).forEach((key) => {
```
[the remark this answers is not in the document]
annasvst: After looking at the intire code, I'd recommend to remove this part completely, it's not needed with react-hook-form.

annasvst (changes requested):

JS/Lesson34/onder/my-app/src/components/NewRecipe/index.jsx
```diff
@@ -1,77 +1,98 @@
...
 
-//   return [ myData, deleteMyData ];
-// }
+  const strCategory = watch('strCategory');
 
-export const NewRecipe = () => {
-  // const [data,  setData] = useMyHook();
+  const onSubmit = (data) => {
+   
+    const ingredientsAndMeasurements = data.strIngredientsAndMeasurements.split(',').map(item => item.trim());
```
annasvst, line 25: The onSubmit function should look like this in the end:
```
	const onSubmit = (data) => {
		const ingredientsAndMeasuresData = getIngredientsAndMeasurements(data);

// Filter data to remove `strIngredientsAndMeasurements` that we have in the form as it's not compatible with API data
		const filteredData = Object.keys(data)
			.filter((key) => key !== 'strIngredientsAndMeasurements')
			.reduce((obj, key) => {
				obj[key] = data[key];
				return obj;
			}, {});

		console.log({
			...filteredData,
			...ingredientsAndMeasuresData,
		});
	};
```

It uses this function (added right outside the component not to pollute the component code): 
```
function getIngredientsAndMeasurements(formData) {
  const ingredientsAndMeasurementsArray = formData.strIngredientsAndMeasurements
      .split(',')
      .map(item => item.trim().split('-').map(part => part.trim()));

  const ingredientsObj = {};
  const measurementsObj = {};

  ingredientsAndMeasurementsArray.forEach(([ingredient, measurement], index) => {
      ingredientsObj[`strIngredient${index + 1}`] = ingredient;
      measurementsObj[`strMeasure${index + 1}`] = measurement;
  });

  return { ...ingredientsObj, ...measurementsObj };
}
```

This function converts {..., "strIngredientsAndMeasurements": "Flour - 100gr, milk - 50ml"} to {..., "strIngredient1": "Flour", "strIngredient2": "milk", "strMeasure1": "100gr", "strMeasure2": "50ml"}.

In real life event more validation would be required but it's ok not to do it right now fort his assignment.

annasvst (approved):

PR #133, opened 2023-11-30
tugbaLesson34-JS

TugbaMartin:


annasvst (changes requested):
Hi Tuğba, this is a good step forward but the pull request requires a bit more work :) 
If you look at the structure of this file `JS/Lesson34/tugba/my-app/src/data.json`, you'll see other fields that need to be provided by the user. E.g. strMealThumb, strTags, strYoutube, strMeasure1 - strMeasure20, strSource, strImageSource, strCreativeCommonsConfirmed. 

Make sure that these fields are not required so the user is not forced to enter value in there. 
Same for the ingredients, a user should ot be force to enter all 20 ingredients. 

Also, when you have multiple similar values like for ingredients here, you don't want to have it as 20 separate input fields in the form. Normally, we'd handle it with a more complex solution where it would be possible to add new input fields if needed for ingredients and measures. However, not to complicate it now you can have a `textarea` input and ask a user to input all ingredients separated by a coma (e.g. "milk, eggs, water, ..."). Then, when you consume the data from the form, you'd need to split the value of `textarea` and add assign it to different `strIngredient` fields.

PR #134, opened 2023-11-30
SAVAS-Post page was constructed.

savaspamuk:


JS/Lesson26/savas/script.js
```diff
@@ -6,9 +6,110 @@ p (.post-body)
...
+        json.forEach(element => {
+            posts += `<div class="post">
+                <h2 class="post-title"><span>Post ${element.id}</span> - ${element.title}</h2>
+                <p class="post-body">${element.body}</p>
+                <button class="button--edit" data-id="${element.id}">Edit Post</button>
+                <button class="button--danger" data-id="${element.id}">Delete Post</button>
+            </div>`;
+        });
+
+        document.getElementById("posts-container").innerHTML = posts;
```
annasvst, line 56: innerHTML is an "unsafe" method. Try to create elements one by one and append them to each other instead. This is a longer approach but it is a safer choice.
E.g.
const heading = document.createElement('h2');
...

Actually,  `<span>Post ${element.id}</span> - ${element.title}` this is the only code that I'd use innerHTML for:
heading.innerHTML = `<span>Post ${element.id}</span> - ${element.title}`;

For other things I'd set attributes and innerText one by one.

annasvst (commented):
Looks good Savaş! I've added comments for things to improve. I'd also change the structure a bit:
main project folder:
- index.html
- styles.css
- script.js
- /create-post folder
  - index.html (inside create-post folder)
  - script.js (inside create-post folder)
- /edit-post folder
  - index.html (inside edit-post folder)
  - script.js (inside edit-post folder)

This way you'd have all the JS code split into different files and you will only need to add JS code that is actually needed in that particular file.

JS/Lesson26/savas/script.js
```diff
@@ -6,9 +6,110 @@ p (.post-body)
...
 
 const URL = 'https://jsonplaceholder.typicode.com/posts';
+
+document.addEventListener("DOMContentLoaded", function() {
+    const getPostsButton = document.getElementById("get-posts");
+    if (getPostsButton) {
+      getPostsButton.addEventListener("click", fetchPosts);
+    }
+  
+    const createPostButton = document.getElementById("create-a-post");
```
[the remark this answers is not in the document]
savaspamuk: I don't know why I have such a perception. It's like I have to define everything. I'm writing extra code for nothing, right :)

JS/Lesson26/savas/script.js
savaspamuk: Anna, I don't understand why you wrote this comment. For the id, title and body of the elements above? (line between 33 and 38)

JS/Lesson26/savas/script-for-edit.js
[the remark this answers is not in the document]
savaspamuk: I did something , but get 404 error for 'https://jsonplaceholder.typicode.com/posts/null'

JS/Lesson26/savas/create-a-post.html
```diff
@@ -0,0 +1,27 @@
...
+    <h3>Create a post</h3>
+    <div id="form-container" class="posts-container">
+        <form id="post-form">
+            <label for="new-title">Title:</label>
+            <input type="text" id="new-title" name="new-title" class="new-title">
+            <label for="new-body">Body:</label>
+            <textarea id="new-body" name="new-body" class="new-body"></textarea>
+            <button id="submit" class="button--success">Submit</button>
+        </form>
+        <a href="index.html"><button  class="button--success">Home Page</button></a>
```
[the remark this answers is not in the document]
savaspamuk: I applied the first step, but for some reason I could not activate the 'home' button. I tried again and it works. Really interesting :)

JS/Lesson26/savas/script.js
annasvst: This comment is about this code:
```
<div class="post">
                <h2 class="post-title"><span>Post ${element.id}</span> - ${element.title}</h2>
                <p class="post-body">${element.body}</p>
                <button class="button--edit" data-id="${element.id}">Edit Post</button>
                <button class="button--danger" data-id="${element.id}">Delete Post</button>
            </div>
```

Instead of adding these elements as innerHTML, it's better to create them as I mentioned above

JS/Lesson26/savas/script-for-edit.js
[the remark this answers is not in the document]
annasvst: The issue with request is that instead of postId you got 'null': https://jsonplaceholder.typicode.com/posts/**null**.

What you need to do in this case is to add console.log statements in different places around you code and try to understand why by the time this request is made the value of `postId` becomes `null`.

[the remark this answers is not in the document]
savaspamuk: I understand that the error is related to getting the ID of the post to be edited. But I couldn't pull the id successfully.

JS/Lesson26/savas/edit-post-folder/index.html
```diff
@@ -0,0 +1,27 @@
...
+  <title>Posts</title>
+  <link rel="stylesheet" type="text/css" href="styles.css">
+</head>
+<body>
+  <div class="container">
+    <h1>Posts</h1>
+    <hr>
+    <h3>Edit The Post "<span id="edited-post"></span>"</h3>
+    <div id="form-container" class="posts-container">
+        <form id="edit-form">
```
annasvst, line 13: I finally noticed the issue with this page. You need to select the form element in JS, not the button. 
When you select the form element, then you can use "submit" event type for all possible ways of submitting a form. When you select the button, you need to use "onclick" event instead and then cover other methods that can be used to submit the form. See the comment in JS code as well for how to fix the current issue.

JS/Lesson26/savas/edit-post-folder/script.js
```diff
@@ -0,0 +1,45 @@
...
+    fetch(`https://jsonplaceholder.typicode.com/posts/${postId}`)
+    .then(response => response.json())
+    .then(json => {
+        document.getElementById("edited-post").innerText = postId;
+        document.getElementById("edit-title").value = json.title;
+        document.getElementById("edit-body").value = json.body;
+    })
+    .catch(err => console.log(err));
+
+    const form = document.getElementById("edit-post-form");
```
annasvst, line 16:
```suggestion
    const form = document.getElementById("edit-form");
```

JS/Lesson26/savas/edit-post-folder/index.html
savaspamuk: Wow, the codes are quickly getting very complex and I'm getting confused about what I'm doing where :)

JS/Lesson26/savas/edit-post-folder/script.js
savaspamuk: That's it! I couldn't find you "minor error" :)
It works Anna, thank you!

JS/Lesson26/savas/script.js
```diff
@@ -6,9 +6,96 @@ p (.post-body)
...
+    document.getElementById("create-a-post").addEventListener("click", function() {
+        window.location.href = "create-post/index.html";
+      });
+  
+    document.getElementById("clear-posts").addEventListener("click", function() {
+        const postsContainer = document.getElementById("posts-container");
+        postsContainer.innerHTML = "";
+      });
+  
+    document.getElementById("submit").addEventListener("click", createANewPost);
```
annasvst, line 23: You need to delete this line since this JS file is not attached to the HTML file containing the create a new post form any more. The `Uncaught TypeError: Cannot read properties of null (reading 'addEventListener')
    at HTMLDocument.<anonymous> (script.js:23:38)` says that the element with id 'submit' is not found.

savaspamuk: Ok, I see. Now, we don't get any error message :)

PR #135, opened 2023-12-06
SAVAS-Lesson 30 Exercises were finished.

savaspamuk:


JS/Lesson30/savas/my-app/src/components/Ex6/index.jsx
```diff
@@ -0,0 +1,14 @@
...
+
+export const Ex6 = () => {
+
+    function handleClick (event) {
+        event.preventDefault();
+        console.log("Button Clicked!")
+    }
+
+    return (
+        <button onClick={handleClick}>Prevent Default</button>
```
annasvst, line 12: in web development, when we say 'link' element we mean <a> element. Clicking on the button would not have the same effect as clicking on the link. The default behaviour of the link click would be redirection so we want to prevent it here.

JS/Lesson30/savas/my-app/src/components/Ex9/index.jsx
```diff
@@ -0,0 +1,40 @@
+// ## Exercise 9: Multiple Buttons
+// Create a component with multiple buttons. Change the background color of the component based on which button is clicked.
+
+export const Ex9 = () => {
+
+    function handleClickYellow () {
+        document.getElementById('root').style.backgroundColor = "yellow";
```
annasvst, line 7: When using React, we normally do not select elements as we did in pure JS (`document.getElementById('root')`).
Since the background that you want to change is not accessible from here, you'd normally use state management on the application level here (Redux or ContextApi). 
No need to change anything in this exercise right now but avoid using pure JS element selector going forward. :)

annasvst (changes requested):
Well done Savaş, I've left a few comments, please take a look when you have time!

JS/Lesson30/savas/my-app/src/components/Ex10/index.jsx
```diff
@@ -0,0 +1,13 @@
...
+// Implement an event handler that triggers when a button is double-clicked. Display an alert indicating that the button was double-clicked.
+
+export const Ex10 = () => {
+
+    function handleDoubleClick () {
+        alert('The button was double-clicked.');
+    }
+
+    return (
+        <button onDoubleClick={handleDoubleClick}>Prevent Default</button>
```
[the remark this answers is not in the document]
savaspamuk: There are two "Prevent Default" button in the page and this codes is for second button. I changed the button name as "Double Click Button".

JS/Lesson30/savas/my-app/src/components/Ex6/index.jsx
savaspamuk: Ouch, I misunderstand it. I changed the button element with a link element.

JS/Lesson30/savas/my-app/src/components/Ex9/index.jsx
savaspamuk: Ok, I understood :)

JS/Lesson30/savas/my-app/src/components/Ex10/index.jsx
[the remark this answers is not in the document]
annasvst: What I mean is that you cannot use `onDoubleClick={handleDoubleClick}`. You need to use `onClick={handleDoubleClick}` and insider `handleDoubleClick` function you need to check `event.detail` to see how many times the button was clicked.

JS/Lesson30/savas/my-app/src/components/Ex6/index.jsx
annasvst: Should be like this: 
```suggestion
        <a href='https://https://google.com' onClick={handleClick}>Link to Google</a>
```

savaspamuk: Anna, I actually made this change, but this comment seems unchanged in the section we added. VS Code gives an alert like "Editor is read-only because the file system of the file is read-only". I clicked the Ex6 file, I can change the codes, but your comment doesn't seem there.

annasvst (approved):
Approved, looks good!

PR #137, opened 2023-12-12
SAVAS-Lesson 32 Exercise was completed

savaspamuk:


JS/Lesson32/savas/my-app/src/App.jsx
```diff
@@ -3,13 +3,15 @@ import { useEffect, useState } from 'react';
 import {BD_SEARCH_BASE_URL} from './urls.js';
 import {RecipeList} from './components/RecipeList';
 import {SearchBar} from './components/SearchBar';
+import { RecipeDetail } from './components/RecipeDetail/index.jsx';
```
annasvst, line 6: When a file in a folder is named 'index.js' or 'index.jsx', you don't need to explicitly write it in the import statement. Instead, you can write like this:
```suggestion
import { RecipeDetail } from './components/RecipeDetail';
```

JS/Lesson32/savas/my-app/src/App.css
```diff
@@ -0,0 +1,3 @@
+div.recipe-list{
```
annasvst, line 1: Since this code is related to the RecipeList component, it's better to add a `styles.css` file in the RecipeList folder and import it to the `RecipeList/index.jsx` file.  This way you project will be cleaner and easier to maintain in long term.

JS/Lesson32/savas/my-app/src/components/RecipeDetail/index.jsx
```diff
@@ -0,0 +1,31 @@
...
+export const RecipeDetail = ({recipe}) => {
+  return (
+    
+    <div>
+        <h2>{recipe.strMeal}</h2>
+        <h3>Country: {recipe.strArea}</h3>
+        <a href={recipe.strSource}>Original Recipe Link</a>
+        <h3>Ingredients:</h3>
+        <ul>
+            <li>{recipe.strIngredient1}</li>
```
annasvst, line 12: You have to repeat yourself a lot when creating `li` elements manually. Instead, try to automate it. Every recipe has `strIngredient` with indexes from 1 to 20 so you'd need to write some JS code that would get all of this 20 values and then display them here in the component. 
It would help a lot if you try to get these 20 values and put the in an array variable first. Then, inside  `ul`, just map through the array and display `li` for each element in array.

annasvst (changes requested):
Hi Savaş, the code works but could be improved a bit, I've added comments about it. Try to optimize RecipeDetail component as suggested. :)

JS/Lesson32/savas/my-app/src/App.css
savaspamuk: Anna, this is actually not a style related to RecipeList. I want to use it to put the recipe list and details side by side. Thus, I think to change the className as "content". Is it appropriate?

annasvst: I see now. It would be better to use a name "container" :)

savaspamuk: Actually, the parent div's class is "container", so I put the name as "content". In that case, is "sub-container" appropirate?

annasvst: I'd say sub-container is not a common name to use but let it be in this case :)

annasvst (approved):
Approved now :)

JS/Lesson32/savas/my-app/src/App.css
savaspamuk: Ok, thank you :)

PR #143, opened 2024-01-15
Lesson34-completed

TugbaMartin:


JS/Lesson34/tugba/my-app/src/components/NewRecipe/index.jsx
```diff
@@ -1,77 +1,99 @@
...
 
-//   return [ myData, deleteMyData ];
-// }
+  const strCategory = watch('strCategory');
 
-export const NewRecipe = () => {
-  // const [data,  setData] = useMyHook();
+  const onSubmit = (data) => {
+   
+    const ingredientsAndMeasurements = data.strIngredientsAndMeasurements.split(',').map(item => item.trim());
```
annasvst, line 25: Well done separating each individual ingredient - measure pairs! Now each data item looks like "Flour - 1dl". You need to convert it to the format that the API supports, see example here: https://www.themealdb.com/api/json/v1/1/search.php?s=Arrabiata
For each ingredient you'll need to use 'strIngredient1' property where number is a number of the ingredient,  for the corresponding measure you'll use 'strMeasure1' (strMeasure2, strMeasure3, etc.) property.

PR #144, opened 2024-01-24
OnderJS18New

ondermartin:


JS/Lesson18/onder/recursion.js
```diff
@@ -13,130 +13,316 @@
...
+	const result = removeSpecified(sentenceEx4, charToRemove);
+	console.log(result);
+	
+	console.log("================================")
+	// 6. Write a recursive function to check if an array includes a specific value.
+	console.log("Example6: ");
+	const numbers = [1, 2, 3, 4, 5];
+	const valueToCheck = 3;
+	
+	function removeSpecified(sentence, charToRemove){
```
annasvst: There's a function with name `removeSpecified` declared for the exercise 5, you need to use a different name here as we cannot declare variables with the same name in the same scope in JavaScript.

annasvst: Also, the function should check if array includes a specific value or not (= return true or false) but the current solution does not do that.

annasvst (changes requested):
Ex. 6 requires more work but otherwise looks good.

PR #152, opened 2024-02-22
Onder-lesson35-36-cont-correction

ondermartin:


JS/Lesson35/onder/my-app/src/App.jsx
```diff
@@ -1,38 +1,53 @@
...
+		<>
+			<CssBaseline />
+			<Navbar />
+			<Container component='main' maxWidth='lg'>
+				<Grid container spacing={3}>
+					<Grid item xs={12}>
+						<SearchBar setRecipes={setRecipes} />
+					</Grid>
+					<Grid item xs={12}>
+						<Routes>
```
annasvst, line 33: You have Routes directly added here inside <Routes> component and also in the external AppRouter.jsx file. Using an external file for defining routes would make your code cleaner but in any way you need to pick one of the approaches onle.

JS/Lesson35/onder/my-app/src/App.jsx
```diff
@@ -1,38 +1,53 @@
...
-      <AppRouter recipes={recipes} setRecipes={setRecipes} />
-			
-		</div>
+		<>
+			<CssBaseline />
+			<Navbar />
+			<Container component='main' maxWidth='lg'>
+				<Grid container spacing={3}>
+					<Grid item xs={12}>
+						<SearchBar setRecipes={setRecipes} />
```
annasvst, line 30: I'd recommend moving SearchBar component somewhere else (e.g. to RecipeList component), otherwise it shows up on every page in the application.

JS/Lesson35/onder/my-app/src/components/UserPage/index.jsx
```diff
@@ -1,3 +1,19 @@
-export const UserPage = () => {
-  return (<div>User page</div>)
-};
\ No newline at end of file
+import React from "react";
+import { Typography, Paper, Container } from "@mui/material";
+
+const UserPage = () => {
```
annasvst, line 4: Since you're currently using route defined directly in the App.jsx component, UserPage is never displayed. You need to add a missing route for it similar to this: `<Route path="/user-settings" element={<UserPage />} />` to whereever you define routes.

JS/Lesson35/onder/my-app/src/components/SearchBar/index.jsx
```diff
@@ -1,32 +1,30 @@
...
+    <form onSubmit={handleSubmit}>
+      <input onChange={(event) => setValue(event.target.value)} value={value} type="text" />
+      <button type="submit">Search</button>
+    </form>
+  );
+};
 
-};
\ No newline at end of file
+export default SearchBar;
```
annasvst, line 30: There's a number of benefits of using named exports (`export const SearchBar = () => ....`) instead of default exports in react (`export default SearchBar;`).  
1. Named exports make it easier to understand what you're importing from a module. There's no guessing about what a file exports or having to peek inside to see the default export. This clarity is especially handy in larger projects where you're juggling loads of components.
2. With default exports, every importer can name the import whatever they like. This flexibility can lead to inconsistency across your project. For instance, one file could import Button as ButtonComponent and another as Btn. Named exports enforce consistent naming, which makes your codebase easier to understand at a glance.
3. Suppose you've got a component or function that needs a name change. If it's a default export, you'll have to manually update the name everywhere it's imported. With named exports, modern IDEs and text editors can automatically update these imports for you. Plus, if you're exporting multiple values from a single module, it's easier to add, remove, or modify exports without affecting the existing import statements in other files.
4. When it comes to optimizing your bundle size, named exports play nicer with tree shaking than default exports. Tree shaking is the process used by bundlers like Webpack to eliminate unused code from the final bundle. Named exports make it easier for these tools to determine which parts of a module are used and which aren't, leading to potentially smaller bundles.

annasvst (changes requested):

PR #153, opened 2024-03-15
Onder-Lesson35-36-38 final/completed 

ondermartin:


JS/Lesson35/onder/my-app/src/App.jsx
```diff
@@ -1,38 +1,37 @@
...
-	);
+const App = () => {
+  return (
+    <GlobalProvider>
+      <CssBaseline />
+      <Navbar />
+      <Container component="main" maxWidth="lg">
+        <Grid container spacing={3}>
+          <Grid item xs={12}>
+            <Routes>
```
annasvst, line 21: This works, however in a larger applications with more routes you'd benefit from extracting routes to a separate file (like AppRouter in this project) so that you don't mix it with component code. 

Since AppRouter.jsx file is not used right now, remove it.

JS/Lesson35/onder/my-app/src/AppRouter.jsx
annasvst, line 1: Remove this file, it's not used

JS/Lesson35/onder/my-app/src/ParentComponent.jsx
annasvst, line 1: I don't think this component is used anywhere, remove :)

JS/Lesson35/onder/my-app/src/components/About/index.jsx
annasvst, line 1: No need to change right now but in a real world project if you use a library like MUI for your project, it's better to use components from it everywhere for consistency.

JS/Lesson35/onder/my-app/src/components/Auth/signIn.jsx
```diff
@@ -0,0 +1,95 @@
+import React, { useState } from "react";
+import { Link } from "react-router-dom";
+import { Button, TextField, Typography, Grid, Paper } from "@mui/material";
+
+const SignIn = () => {
```
annasvst, line 5: It would be good to store user data in the Context so that when user fills a sign in / sing up form, we consider the user logged in and show pages like 'create new recipe' and 'settings'. When user click log out button, then the user state is changed to logged out and the mentioned above pages are not shown in the navbar.

JS/Lesson35/onder/my-app/src/components/Auth/signIn.jsx
```diff
@@ -0,0 +1,95 @@
...
+          <Typography variant="body2" sx={{ mt: 2 }}>
+            Don't have an account? <Link to="/signup">Sign Up</Link>
+          </Typography>
+        </Paper>
+      </Grid>
+    </Grid>
+  );
+};
+
+const SignUp = () => {
```
annasvst, line 50: Nit: Usually, it's always a good idea to keep one component per file to increase readability of your code.

JS/Lesson35/onder/my-app/src/components/Navbar/index.jsx
```diff
@@ -0,0 +1,68 @@
+import React, { useContext, useState } from "react";
+import { NavLink } from "react-router-dom";
+import "./styles.css";
+import { GlobalContext } from "../../Context/GlobalState";
+
+const Navbar = () => {
```
annasvst, line 6: Add Context for the user state to the application, then in navbar hide an show certain pages depending on weather user is logged in or not.

JS/Lesson35/onder/my-app/src/components/Navbar/index.jsx
```diff
@@ -0,0 +1,68 @@
...
+
+  const handleSubmit = async (event) => {
+    event.preventDefault();
+    try {
+      const response = await fetch(
+        `https://www.themealdb.com/api/json/v1/1/search.php?s=${searchQuery}`
+      );
+      const data = await response.json();
+      if (data.meals) {
+        setRecipesnew(data.meals);
```
annasvst, line 18: This approach always removes whatever recipe data we had before which may not always be ideal. If you want to preserve previous data, you'd need to create context in a way we did in the lesson (using reducer function would be better in this case). Otherwise, you can use Redux as well but also with reducer function.

JS/Lesson35/onder/my-app/src/components/NewRecipe/index.jsx
annasvst, line 1: New recipe component is missing a lot of functionality right now like fields that are needed to fill data similar to recipes that we get from API. Also, on Submit event we would want to add the newly create recipe to whatever recipes that we already store in the context so that you can also view newly created recipes on the main page.

JS/Lesson35/onder/my-app/src/components/RecipeList/index.jsx
```diff
@@ -1,18 +1,81 @@
-import { Link } from 'react-router-dom';
+// RecipeList.jsx
```
annasvst, line 1:
```suggestion
```

JS/Lesson35/onder/my-app/src/components/RecipeList/index.jsx
```diff
@@ -1,18 +1,81 @@
...
+        console.error("Error fetching recipes:", error);
+      }
+    };
+
+    handleSubmit();
+  }, [setRecipesnew]);
+
+  return (
+    <div className="recipe-list">
+      <Typography></Typography>
```
annasvst, line 41:
```suggestion
```

JS/Lesson35/onder/my-app/src/components/RecipeList/index.jsx
```diff
@@ -1,18 +1,81 @@
...
+                </CardContent>
+              </Card>
+            </ListItem>
+          ))}
+      </List>
+    </div>
+  );
 };
+
+export default RecipeList;
```
annasvst, line 81: Nit: When working on a project, try to use either default exports of named exports, it's better to keep things consistent :)

PR #158, opened 2024-04-24
Lesson35-38-Finall

ondermartin:


JS/Lesson35/onder/my-app/src/App.jsx
```diff
@@ -1,38 +1,39 @@
...
+      <Navbar />
+      <Container component="main" maxWidth="lg">
+        <Grid container spacing={3}>
+          <Grid item xs={12}>
+            <Routes>
+              <Route path="/recipe/:id" element={<RecipeDetail />} />
+              <Route path="/create-recipe" element={<NewRecipe />} />
+              <Route path="/" element={<RecipeList />} />
+              <Route path="/about" element={<About />} />
+              <Route path="/LogIn" element={<LogIn />} />
```
annasvst: Keep routes consistent, change 'LogIn' to 'log-in'

JS/Lesson35/onder/my-app/src/AppRouter.jsx
annasvst, line 1: I think I commented on this before but since you have routes defined directly in the App.jsx file, you don't need to keep this file too. It's a bad practice to keep unused code in the project.

JS/Lesson35/onder/my-app/src/components/Auth/AuthProvider.js
```diff
@@ -0,0 +1,23 @@
+import React, { createContext, useState, useContext } from "react";
+
+const AuthContext = createContext();
+
+export const AuthProvider = ({ children }) => {
+  const [isLoggedIn, setIsLoggedIn] = useState(false);
```
annasvst: Store email and password on the user in the context as well.

Try to use useReducer hook for AuthProvider, it's a better choice as it allow to define explicitly how to handle a certain data change.

JS/Lesson35/onder/my-app/src/components/Auth/LogIn.jsx
```diff
@@ -0,0 +1,70 @@
...
+            label="Password"
+            type="password"
+            margin="normal"
+            value={password}
+            onChange={(e) => setPassword(e.target.value)}
+          />
+          <Button variant="contained" onClick={handleLogIn} sx={{ mt: 2 }}>
+            LogIn
+          </Button>
+
```
annasvst: I don't think anything on lines 49-62 is really needed. Log in form should only be shown to users that are not logged in. If a logged in user navigates to the login page, it should be redirected to, e.g. main page.

JS/Lesson35/onder/my-app/src/components/Navbar/index.jsx
annasvst, line 1: It would be good to show some pages based on whether user is logged in or not. E.g. log in page should not be shown to a logged in user. And add recipe or user settings should not be shown to a user that is not logged in.

annasvst (changes requested):

